PR #177, opened 2026-06-16
fix(cli): Resume must verify checkpointed outputs exist

### LINKED ISSUE #176: Resume silently drops batches whose outputs are missing
When restarting an interrupted run, gwmock loaded the checkpoint, logged **"Loaded checkpoint: 3 batches already completed"**, skipped batches 0/1/2, ran 3/4/5, and reported **"Simulation completed successfully."** — yet the output directory contained **5 files per detector instead of 6**. The missing file is **batch 0**: both its `.gwf` and its `orchestration-0.metadata.json` are absent, while batches 1–5 are all present and internally consistent (metadata `batch_index` ↔ filename `start_time` line up exactly). No error was raised.

fcireddu:
Close #176 

A 6-batch noise run that was interrupted and resumed logged "3 batches already completed", skipped batches 0–2, and reported "Simulation completed successfully" — but produced **5 files per detector instead of 6** (batch 0's `.gwf` and metadata were gone). Resume decided what to skip purely from the checkpoint's `completed_batch_indices`, **never checking that those batches' outputs existed on disk**. A related defect: the output file is written before the checkpoint is committed, so an interrupt leaves an orphan output that aborts the next `overwrite=False` resume with `FileExistsError`.

**Change** (gwmock only): `execute_plan` now reconciles the checkpoint against the filesystem before skipping anything.

- `reconcile_completed_batches` validates each checkpointed batch in order against its
  recorded `outputs[].path` (from `{sim}-{idx}.metadata.json`) and keeps only the valid
  contiguous prefix. The check is simulator-agnostic — it covers noise, signal, full
  signal+noise, and glitch runs (glitches are baked into the noise `.gwf`).
- If any checkpointed batch's output is missing, the checkpoint is discarded and the run
  regenerates from the start (the checkpoint holds only the tail state, so an interior
  batch can't be restored in isolation), with a clear warning. The silent-success
  failure mode is gone.
- On resume, batches being re-run overwrite stale orphan/partial outputs regardless of
  `--overwrite`, so an interrupted-mid-write batch resumes without `FileExistsError`
  (covers both the noise guard and base `Simulator.save_data`). A normal first run still
  honours `overwrite=False`.

**Tests**: added `TestCheckpointReconciliation` (4 cases: prefix truncation on missing output/metadata, regenerate-on-missing-output, orphan tolerance without `--overwrite`, happy-path skip). Full suite: **537 passed, 23 skipped**. Reproduced the exact user symptom (5/6 files) and confirmed the fix restores all 6.

## Release Notes

* **New Features**
  * Added filesystem-aware checkpoint reconciliation that validates previously-recorded outputs exist on disk before skipping batches, preventing unnecessary re-execution of incomplete work during resumption.

* **Bug Fixes**
  * Fixed inconsistent overwrite behavior when resuming simulations with partial outputs from prior runs, ensuring correct handling across the entire batch execution lifecycle.

### DIFF
--- a/src/gwmock/cli/simulate_utils.py
+++ b/src/gwmock/cli/simulate_utils.py
@@ -830,6 +830,97 @@ def validate_plan(plan: SimulationPlan) -> None:
     logger.info("Simulation plan validation completed successfully")
 
 
+def _resolve_recorded_output_paths(metadata: dict[str, Any], working_directory: str | None) -> list[Path]:
+    """Resolve the output file paths recorded in a batch's metadata.
+
+    Paths in ``outputs[].path`` are stored relative to the working directory (see
+    ``_to_path_string``); resolve them back against it so existence can be checked.
+    """
+    base = Path(working_directory) if working_directory else None
+    paths: list[Path] = []
+    for record in metadata.get("outputs", []) or []:
+        raw = record.get("path") if isinstance(record, dict) else None
+        if not raw:
+            continue
+        path = Path(raw)
+        if base is not None and not path.is_absolute():
+            path = base / path
+        paths.append(path)
+    return paths
+
+
+def _batch_outputs_present(batch: SimulationBatch, metadata_directory: Path) -> bool | None:
+    """Return whether the outputs recorded for ``batch`` all exist on disk.
+
+    Returns:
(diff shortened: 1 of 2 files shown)


fcireddu:
@coderabbitai review

PR #226, opened 2026-07-04
fix: accept literal file_name lists in save_data for metadata replay

isaac-cf-wong:
## Summary

`gwmock simulate <metadata.json>` failed for any run whose **signal** section used a multi-detector template in `file_name` (e.g. `E-{{ detectors }}_BBH-{{ start_time }}-{{ duration }}.gwf` — the canonical BBH workflow from the bundled examples):

- **signal-only** replay → `ValueError: Signal output files must end with .gwf, .hdf5, .h5, .npy, or .txt.`
- **signal+noise** replay → noise frames written, signal batch fails and is retried, retry collides with the just-written noise frames → `FileExistsError: Noise adapter output(s) already exist …`

**Root cause:** the metadata sidecar stores the *rendered per-detector list* (`"file_name": ["E-ET1_…gwf", "E-ET2_…gwf"]`) rather than the template. On replay `Simulator.save_data` did `str(file_name)` on that list, producing a single bogus path whose suffix `.gwf']` fails `_infer_signal_output_format`.

**Fix:** `save_data` now resolves list/tuple/ndarray `file_name` inputs entry by entry and hands `_save_data` a 1-D object array of paths — the same already-expanded-literal handling the noise adapter gained for per-batch replay in ISS-016 (`_active_noise_file_name`). `_save_data` already supported per-detector array writes, so no change below `save_data` is needed. Noise-only and single-file signal replay (e.g. SGWB `{{ counter }}.hdf5`) were unaffected and stay unchanged.

## Tests

- `tests/simulator/test_base_simulator.py`: unit regression — `save_data` with a literal file-name list writes one file per entry with `output_directory` prepended; the `overwrite=False` existence check still fires per entry.
- `tests/cli/test_cli_orchestration.py::test_simulate_reproduces_metadata_with_multi_detector_signal_template`: end-to-end — simulate with a `{{ detectors }}` signal template (2 detectors × 2 batches), assert the sidecar records the rendered list, then replay both metadata files in a fresh directory and assert all signal + noise outputs appear.

Both tests **fail on `main`** (verified by reverting the src change) and pass with the fix. Full suite: 566 passed, 23 skipped. Also verified against the real-world repro that surfaced this (Leuven workshop prep, released 0.8.2): the previously failing BBH signal+noise and signal-only metadata replays now succeed with **bit-identical** strain vs the original run.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Simulator output saving now supports multiple detector-specific file name templates, allowing per-detector paths to be generated and saved in one run.
  * Replay and orchestration flows now preserve and reuse expanded per-detector output file lists correctly.

* **Bug Fixes**
  * Improved file overwrite checks so existing outputs in multi-file saves are detected and handled correctly.
  * Metadata now records expanded output paths instead of the original template when multiple detector files are produced.

### DIFF
--- a/src/gwmock/simulator/base.py
+++ b/src/gwmock/simulator/base.py
@@ -281,7 +281,7 @@ class Simulator(ABC):
     def save_data(
         self,
         data: Any,
-        file_name: str | Path,
+        file_name: str | Path | list[str] | np.ndarray[Any, np.dtype[np.object_]],
         output_directory: str | Path | None = None,
         overwrite: bool = False,
         **kwargs: Any,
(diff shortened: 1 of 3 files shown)


isaac-cf-wong:
@coderabbitai review

PR #254, opened 2026-07-13
feat: add signal.waveform-options for extra waveform options

isaac-cf-wong:
Adds a `signal.waveform-options` config field, forwarded to gwmock-signal as its `waveform_arguments` parameter (LAL dictionary entries such as ModeArray) — distinct from the existing `waveform-arguments`, which flat-merges fixed physical parameters into the per-event set.

- Guards: specifying options both ways raises; an old gwmock-signal without `waveform_arguments` support raises an upgrade error instead of silently dropping the options through the unsupported-parameter retry path.
- Recorded in orchestration metadata.
- Requires gwmock-signal >= 0.11.0 (waveform_arguments across all backends: Leuven-Gravity-Institute/gwmock-signal#305, #306, #308, #309, #311).

Verified end-to-end against the real 0.11.0 backend: `waveform-options` `{ModeArray: [(2,2),(2,-2)]}` measurably changes the projected strain. Full signal + cli suites pass (399).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added support for configuring waveform options in signal generation.
  * Waveform options can now be provided through configuration and are applied consistently across signal modes.
  * Added validation for conflicting waveform settings and clearer errors when the installed signal backend lacks required support.

* **Chores**
  * Updated the minimum supported signal adapter version.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -22,7 +22,7 @@ classifiers = [
 requires-python = ">=3.12"
 dynamic = ["version"]
 dependencies = [
-  "gwmock-signal>=0.10.0",
+  "gwmock-signal>=0.11.0",
   "gwmock-noise>=0.7.1",
   "gwmock-pop>=0.11.1",
   "typer>=0.19.0",         # Literal-type CLI options; the SPEC 0 floor (0.13) predates them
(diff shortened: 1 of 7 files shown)


isaac-cf-wong:
@coderabbitai review

PR #270, opened 2026-07-23
feat(reproducibility): replay runs to their resolved dataset versions

isaac-cf-wong:
## Summary

Completes the A+B data-versioning design (gwmock-noise producer side shipped in
v0.9.0). Replay reconstructed each batch only from the **input** config,
discarding all runtime-resolved state — so a run using a HuggingFace-backed
`DeepExtractorGlitch` that was **not** explicitly pinned could not be
reproduced once the upstream dataset moved.

Adds a replayable `resolved_config` metadata layer:

- **Backend contract** — adapters may expose `resolved_config()`. The noise
  adapter retains its glitch models and returns them resolved (pinned to a
  concrete HF commit SHA) and re-serialized; the orchestrator aggregates these
  into an `OrchestrationConfig`-shaped fragment.
- **Producer** — save-time deep-merges that fragment over the template-expanded
  input config into `metadata.resolved_config`, and records a `replayable`
  flag that is `false` (with a warning) when an external-mutable input is not
  pinned to an **immutable full commit SHA** (a branch/tag/`None` does not
  count).
- **Replay** — prefers `resolved_config` over `config`, so an unpinned run
  replays to the exact resources it used; falls back to `config` for legacy
  metadata and warns when replaying a non-replayable run.

Parametric-only runs are unaffected (`resolved_config` is `null`,
`replayable: true`). Signal/population producers can adopt the same contract
later.

## Reproducibility

An unpinned DeepExtractor run now records the concrete commit it used and
refetches exactly that on replay — bit-reproducible for a fixed
(version, config, seed) even as the upstream dataset moves.

## Testing

- New `tests/cli/test_resolved_config_replay.py` (14 tests): adapter
  `resolved_config()`, deep-merge, SHA/symbolic/None pin predicate, build +
  replayable flag + warning, replay preference/fallback, and an end-to-end
  real-DeepExtractor pin with the Hub mocked (no network).
- Full suite: 690 passed, 23 skipped. Ruff clean.
- Reviewed by a second agent; 1 HIGH (symbolic-ref false-positive) + 1 LOW
  (stale models on adapter reuse) found, fixed, re-confirmed closed.

## Dependency

Bumps `gwmock-noise>=0.9.0` for the glitch `resolve()`/`serialize()` contract.

## Summary by CodeRabbit

- **New Features**
  - Added resolved configuration metadata that records runtime-pinned external inputs.
  - Replays now prefer resolved configurations to reproduce runs using the exact versions originally used.
  - Added replayability status to run metadata.

- **Bug Fixes**
  - Runs with mutable or unpinned external inputs are now identified as non-replayable and produce a warning instead of implying bit-for-bit reproducibility.

- **Documentation**
  - Updated reproducibility guidance and metadata examples to explain resolved configurations, replay behavior, and replayability warnings.

### DIFF
--- a/docs/user-guide/reproducibility.md
+++ b/docs/user-guide/reproducibility.md
@@ -18,6 +18,8 @@ Consumers must reject unknown major versions.
         "gwmock_pop": "x.y.z"
     },
     "config": {},
+    "resolved_config": {},
+    "replayable": true,
     "config_sha256": "...",
     "seed": 42,
     "segment_seeds": [123456789, 987654321],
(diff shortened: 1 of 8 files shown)


isaac-cf-wong:
@coderabbitai review

PR #273, opened 2026-07-23
feat(reproducibility): recreate the recorded environment on --isolate

### LINKED ISSUE #11: Add feature to verify version requirements and auto-create isolated environment during reproduction
When reproducing files using metadata, it is important to ensure that the software version used matches the version recorded in the metadata. To improve reproducibility and prevent inconsistent results, the system should verify version requirements and, when mismatched, automatically create a temporary isolated environment with the correct version.

isaac-cf-wong:
## Summary

Fixes #11. Reproducing from metadata could silently run against different
dependency versions than the original run.

- **Record** a full environment freeze — Python version + every installed
  distribution (direct *and* transitive) — in each metadata record (new
  `environment` field; `SCHEMA_VERSION` 1.2.0 → 1.3.0).
- **`gwmock simulate <metadata> --isolate`** reads that freeze, builds a cached,
  isolated **uv** virtualenv pinned to it (matching the recorded Python
  `major.minor`, keyed by the version set so repeated reproductions reuse it),
  and re-execs the reproduction inside it. Runs in place if the current
  environment already matches **exactly** (extra installed packages count as a
  mismatch); warns + runs in place if no environment was recorded (older
  metadata). `--dry-run` is propagated to the isolated child.
- **Without `--isolate`**, reproduction warns when the recorded environment
  differs and points at `--isolate`.
- Concurrent builds of the same environment are serialized with a file lock. uv
  absence or unresolvable recorded versions **fail loudly** rather than run in
  the wrong environment.

## Testing

- New `tests/cli/utils/test_environment.py` (capture/diff/match/key/build/re-exec)
  and `tests/cli/test_simulate_isolate.py` (flag wiring, YAML/mixed metadata,
  match/mismatch, dry-run propagation).
- Full suite: 733 passed, 23 skipped. Ruff clean. End-to-end orchestration run
  asserts the freeze lands in written metadata.
- Reviewed by a second agent: 2 HIGH (dry-run dropped; extra-package match) + 2
  MED (YAML metadata ignored; mixed environments) fixed; build concurrency
  hardened with a file lock.

## Limitation

Recreation needs `uv` and the recorded versions must be resolvable from your
index; a run made with editable/dev installs cannot be recreated this way and
fails loudly rather than running in the wrong environment.

Fixes #11

## Summary by CodeRabbit

* **New Features**
  * Added `--isolate` to reproduce simulations using the recorded Python and package environment.
  * Simulation metadata now records environment details for more precise reproducibility.
  * Added caching and reuse of isolated environments when dependencies match.

* **Documentation**
  * Updated reproducibility guidance and metadata schema to version 1.3.0.
  * Documented isolation requirements, fallback behavior, and environment constraints.

* **Bug Fixes**
  * Added warnings for environment differences during reproduction.

### DIFF
--- a/docs/user-guide/reproducibility.md
+++ b/docs/user-guide/reproducibility.md
@@ -5,12 +5,12 @@
 
 ## Schema
 
-Each record is validated at write time and uses schema version `1.2.0`.
+Each record is validated at write time and uses schema version `1.3.0`.
 Consumers must reject unknown major versions.
 
 ```json
 {
-    "schema_version": "1.2.0",
+    "schema_version": "1.3.0",
     "gwmock_version": "x.y.z",
     "subpackage_versions": {
         "gwmock_signal": "x.y.z",
@@ -62,6 +62,11 @@ Consumers must reject unknown major versions.
         "python": "3.12.x",
         "cpu": "...",
         "git_sha": "..."
+    },
+    "environment": {
+        "python": "3.12.5",
+        "python_implementation": "CPython",
+        "packages": { "numpy": "2.0.0", "gwmock": "x.y.z", "...": "..." }
     }
 }
 ```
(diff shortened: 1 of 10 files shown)


isaac-cf-wong:
@coderabbitai review

PR #286, opened 2026-07-31
feat(signal): add a batched entry point so gwmock can simulate on device

isaac-cf-wong:
## 📝 Summary

**gwmock could not run a GPU simulation at all.** Not slowly — at all. It had no call site for
gwmock-signal's batched path: `cli/adapter_orchestration.py` loops over population events calling
`simulate` one at a time, which is the only shape the LAL and PyCBC backends offer.

`SignalAdapter.simulate_segments` is the batched counterpart — catalogue in as a struct-of-arrays,
assembled segments out — with `events_to_struct_of_arrays` to transpose what the orchestration layer
holds.

It is deliberately a **second** entry point rather than a change to `simulate_stack`, because the
per-event path must keep working unchanged for backends with no batched form.

### The device path always generates with Ripple

Worth stating plainly, because it is the sharpest edge here. Ripple is the only JAX implementation
and therefore the only batchable one, so this path ignores the configured backend. An adapter built
for LAL that runs here gets Ripple's version of the approximant — close agreement, not bit-for-bit.

The adapter cannot detect which backend it holds: `CBCSimulator` keeps its `waveform_backend` inside
a `WaveformFactory` and exposes no attribute for it. What it *can* check is the approximant, so
`_device_approximant()` rejects any approximant Ripple does not implement and lists the alternatives.
A `SEOBNRv4` adapter errors instead of quietly becoming a Phenom waveform. The remaining
implementation difference is documented rather than left unsaid.

### Verified end to end

Against the released gwmock-signal 0.12.0: six BNS events through the bundled `ET-Triangle-Sardinia`
preset give **four assembled segments across three channels, 493,755 non-zero samples, peak strain
3.6e-23**.

### Two dependency changes

The floor moves to `gwmock-signal>=0.12.0`, the first release exporting the device entry points — a
resolver could otherwise legally pick an older 0.11.x, where the import fails.

A **`jax` extra** is added, mirroring `sgwb`. This is a real gap independent of everything else:
gwmock declared no such extra, so `ripplegw` was absent from a standard install and the device path
was unreachable **by installation**, not just by code. Note this installs generic JAX, not a
CUDA-enabled build — GPU deployment instructions are still owed, and are not in this PR.

## ✨ Type of Change

- [x] 🐛 Bug fix (non-breaking change which fixes an issue) — the missing `[jax]` extra
      expected)
- [ ] 📝 Documentation update

Additive. The per-event path is untouched; no existing behaviour changes.

- [x] **Unit Tests:** `uv run pytest` — **750 passed, 23 skipped**. The skips are pre-existing
      legacy-config ones, verified identical on `main`. 15 new tests in
      `tests/signal/test_adapter_batch.py`.
- [x] **Manual Test:** verified against the **released** 0.12.0 from PyPI, not only a local build.
- [x] **Manual Test:** red check — with `adapter.py` reverted to the first commit, all 7 tests added
      in the second commit fail and only the 8 pre-existing ones pass.
- Input-handling tests run without the `[jax]` extra; only those reaching Ripple skip.

## 🏗️ Checklist

Not applicable.

## What the independent review changed

An independent review found two ways this could silently generate a different waveform than the one
configured, and both were real.

`_waveform_model_name` searched `waveform_model`, then `_waveform_model`, then `approximant` on the
backend. `_waveform_model` turned out to be the private backing field of the public `waveform_model`
property, so the chain read the same value twice and then tried a name nothing defines — three
candidates, one real source. Now it reads the public property only, since a backend that grew a
similarly-named attribute for another purpose would otherwise be picked up silently.

**Callable waveform models** are registered under a generated key by the per-event path. That key was
reaching Ripple as if it were an approximant name, so the failure talked about an unsupported
approximant called `__gwmock_custom__...` rather than about the callable. Rejected explicitly now.

**Ragged events now raise.** This reverses the original behaviour and the reasoning behind it. The
docstring said intersecting to the shared keys avoids fabricating physics; it does the opposite — if
one event omits `spin_1z`, the column disappears and the backend default is applied to *every* event,
with nothing said. That is worse than the malformed catalogue it was concealing.

Two more: length validation tested `isinstance(list, tuple)`, so an `ndarray` of the wrong length
bypassed it and failed inside batching where the message no longer names the parameter; and
`earth_rotation` was not forwarded, so it was honoured on the per-event path and ignored here.

The review also caught a test whose name overpromised — it asserted through a failure message, which
passes for either merge order and so could not detect the precedence being backwards, the one thing
it existed to check. It now captures what actually reaches `simulate_cbc_catalogue`.

## Not in this PR

The CLI still cannot reach this. `_simulate()` is not wired to it and no config key selects a compute
backend, so the path exists for API callers but not from a YAML run. That is the next slice — see
`gwmock/device-call-site` and `gwmock/config-compute-backend`.

Also outstanding: per-event provenance on segments, assembly ownership, and CUDA-specific install
documentation.

## Summary by CodeRabbit

- **New Features**
  - Added batched gravitational-wave signal simulation for generating detector segments from event catalogues.
  - Added support for device-accelerated simulation through the optional JAX extra.
  - Added configurable chunking, waveform arguments, Earth-rotation handling, and simulation time ranges.
  - Added validation for event parameters, input shapes, and supported waveform approximants.
- **Compatibility**
  - Updated the signal simulation backend requirement to version 0.12.0.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -22,7 +22,7 @@ classifiers = [
 requires-python = ">=3.12"
 dynamic = ["version"]
 dependencies = [
-  "gwmock-signal>=0.11.2",
+  "gwmock-signal>=0.12.0", # 0.12.0 first exports the on-device entry points simulate_segments needs
   "gwmock-noise>=0.10.0",
   "gwmock-pop>=0.11.4",
   "typer>=0.19.0",         # Literal-type CLI options; the SPEC 0 floor (0.13) predates them
@@ -36,7 +36,10 @@ dependencies = [
 ]
 
 [project.optional-dependencies]
-sgwb = ["gwmock-signal[sgwb]>=0.11.2"]
+sgwb = ["gwmock-signal[sgwb]>=0.12.0"]
+# The on-device (GPU) signal path. Without this extra `SignalAdapter.simulate_segments` raises,
+# because gwmock-signal's batched entry points need JAX and ripple, which the base install omits.
+jax = ["gwmock-signal[jax]>=0.12.0"]
 
 [dependency-groups]
 dev = [
--- a/src/gwmock/signal/adapter.py
+++ b/src/gwmock/signal/adapter.py
@@ -9,6 +9,7 @@ from functools import lru_cache
 from pathlib import Path
 from typing import Any
 
+import numpy as np
 from gwmock_signal import CustomDetector, DetectorStrainStack, Network, resolve_simulator_backend
 
 from gwmock.data.time_series.time_series import TimeSeries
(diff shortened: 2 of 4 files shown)


isaac-cf-wong:
@coderabbitai review

PR #287, opened 2026-07-31
docs(examples): index every example and fix the end-to-end test matrix

isaac-cf-wong:
## 📝 Summary

`examples/` had **no index**, and the 26 configs in it are not the matrix of execution paths they
look like.

Ten differ only in which detector's glitch population they read. Twelve only in the detector
network. Around 22 collapse onto roughly five distinct code paths — while four of the six CLI
commands appear in no example at all, and the output-format branches (`npy`, `txt` for signal;
`npy` for noise) have zero coverage.

This adds two things and no behaviour change.

**`examples/README.md`** indexes all 26: a goal-oriented table for picking one, then a full listing
using `<network>` and `<e1|e2|e3>` placeholders so 26 rows stay browsable rather than becoming a
wall.

**`tests/e2e/matrix.py`** fixes the subset the end-to-end suite will drive — one entry per distinct
code path, not per configuration. Four networks resolve through one `Network` class and `bbh`/`bns`
both resolve to `CBCSimulator`, so those are represented once. Unit tests are what guarantee the
rest follow, and that assumption is recorded as the reason the subset is legitimate.

### Why the drift tests

An index that lists coverage gets _read_ as a statement about coverage, so it is worth enforcing
rather than trusting. Three ways it can lie, one test each: an example missing from the index, an
index row naming an example that is gone, and the documented matrix disagreeing with the matrix in
code.

The third earned its keep immediately — markdown escapes pipes inside tables, so the alternation
placeholder is really `<e1\|e2\|e3>`, and my pattern expansion silently failed to match **nine**
glitch examples. That would have shipped as a confidently-wrong index.

### On BNS being excluded

Two independent grounds. `resolve_simulator_backend` returns the same `CBCSimulator` for `bbh` and
`bns` — they differ in waveform model and tidal parameters, which unit tests own. Separately, BNS is
the expensive case: its inspiral is roughly **160 s at 20 Hz**, so it cannot be shortened to a
test-sized segment without truncating the chirp and simulating something physically wrong.

That 160 s is 0PN scaling from a measured 1033 s at 10 Hz, so treat it as ±20%, not exact. It is the
constraint that decides BNS cannot be in a fast matrix, so the direction matters more than the
digits.

## ✨ Type of Change

      expected)
- [x] 📝 Documentation update

No source changes. Nothing under `src/` is touched.

- [x] **Unit Tests:** `uv run pytest` — **762 passed, 23 skipped**. The skips are pre-existing
      legacy-config ones, verified identical on `main`. 10 new tests in
      `tests/e2e/test_examples_index.py`.
- [x] **Manual Test:** the drift tests were confirmed red before being made to pass — the
      escaped-pipe bug above was found that way, not by inspection.
- [x] **Manual Test:** verified the parser still passes after `prettier` reformatted the tables,
      since it depends on that formatting.

## 🏗️ Checklist

- [x] I have made corresponding changes to the documentation — this PR _is_ the documentation.

Not applicable.

## What this deliberately does not do

**No runner.** Nothing executes the matrix yet; the entries are checked for existence only. The
README says so in a status note rather than claiming coverage that does not exist — writing "the
suite runs these configs" while no runner existed would be precisely the lie the drift tests are
here to prevent.

**No examples deleted.** An earlier plan was to collapse the geometry duplicates. Keeping all 26 and
testing a subset serves users and tests separately, which is better than making the examples worse
to make the tests cheaper.

Still to come, each separately: the `e2e` marker and its own CI job, absorbing
`tests/cli/test_sgwb_end_to_end.py`, the test-time overlay that shortens durations without editing
these files, a determinism probe, and then stored reference values.

## Sequencing

`feat/waveform-backend-selection` is stacked on this branch and adds a
`signal/waveform_backend/ripple` example with its README row and matrix entry. Merging this first
lets that one rebase onto `main` cleanly.

## Summary by CodeRabbit

* **Documentation**
  * Added comprehensive guidance for example configurations, CLI usage, detector networks, testing, overlays, and contribution workflows.

* **Tests**
  * Added end-to-end coverage for selected example configurations and optional dependency scenarios.
  * Added validation ensuring the examples documentation stays complete, accurate, and synchronized with the test matrix.

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -56,16 +56,18 @@ each is runnable without editing.
 The glitch examples are per-detector rather than per-network because each
 detector draws from its own glitch population file.
 
-## Which of these the test suite runs
+## The end-to-end test matrix
 
-A **subset** of these examples is the end-to-end test matrix: the configs the
-suite drives through the real CLI to check a run completes, writes the expected
-files, and reproduces stored reference values.
+A **subset** of these examples is declared as the end-to-end matrix: the configs
+intended to be driven through the real CLI, checking that a run completes,
+writes the expected files, and reproduces stored reference values.
 
-> **Status.** The matrix below is fixed and enforced — `tests/e2e/matrix.py`
-> holds it as data, and a test fails if this table and that file disagree. The
-> runner that executes the configs lands separately; until it does, these
-> entries are checked for existence, not yet run.
+> **Status — declared, not yet executed.** No runner drives these configs today.
+> What is enforced now is that the set is fixed and consistent:
+> `tests/e2e/matrix.py` holds it as data, a test fails if this table and that
+> file disagree, and every entry must name an example that exists. Nothing yet
+> checks that running one reaches the code path claimed for it. The runner lands
+> separately, and this note goes when it does.
 
 The subset is chosen to cover each distinct **code path** exactly once, not
 every configuration. Where two examples differ only in values that flow through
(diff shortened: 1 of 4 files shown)


isaac-cf-wong:
@coderabbitai review

PR #288, opened 2026-07-31
feat(config): select the waveform library from a configuration file

isaac-cf-wong:
## 📝 Summary

PyCBC and ripple are supported by gwmock-signal and were **unreachable from YAML**.
`signal.backend` was already taken — it names the _simulator_ (`bbh`, `bns`, `sgwb`) via
`resolve_simulator_backend` — so there was no key for the waveform library, and LAL was the only
reachable choice.

`signal.waveform-backend` fills that gap: `lal` (default), `pycbc`, `ripple`, `gwsignal`, an entry
point in `gwmock.waveform`, or a `module:Class` reference. Constructor arguments go in
`signal.waveform-backend-arguments`, which is how ripple's `taper_fraction` becomes settable from a
config file for the first time.

### The gap was smaller than it looked

The plumbing already existed. `signal.arguments` is key-normalised and splatted into the simulator
constructor, so `signal.arguments.waveform-backend` _did_ arrive at `CBCSimulator`. It then failed,
because YAML can express a string and `WaveformFactory` wants an instance:

```
AttributeError: 'str' object has no attribute 'available_approximants'
```

So the missing piece was only a name→instance resolver. This adds a `"waveform"` kind to the
**existing** backend resolver rather than a parallel mechanism, which means aliases, entry points and
`module:Class` references all behave as they do for the other kinds — and third-party waveform
backends become pluggable on the same terms.

### Provenance, which this PR would otherwise have broken

Run metadata recorded `waveform_model` but not the library. Since this is the _first_ way to vary the
library, shipping it unrecorded would mean two runs with measurably different strain carrying
identical provenance — bad in a repo built around replay and content hashing. Now recorded as
`orchestration.signal.waveform_backend`, derived from the config so there is one source of truth.

## ✨ Type of Change

      expected)
- [x] 📝 Documentation update

Additive. Omitting the key keeps LAL, and a test pins that.

- [x] **Unit Tests:** `uv run pytest` — **781 passed, 23 skipped** (pre-existing legacy-config
      skips). 19 new tests.
- [x] **Manual Test:** red check — with the three source files reverted, 16 of the new tests fail.
      The two that pass either way assert _unchanged_ behaviour (default stays LAL; a run contains
      signal), so they are guards and correctly non-discriminating.
- [x] **Manual Test:** the shipped example runs end to end — 18,432 non-zero samples, peak
      8.694e-22.
- [x] **Manual Test:** `lal` still resolves with the `ripplegw` import blocked, so a base install is
      unaffected.

## 🏗️ Checklist

Not applicable.

## Measurements

**The key changes the data, not just the object graph.** Same config, `lal` vs `ripple`: peak
**8.385e-22** vs **8.434e-22**, with the merger in the **same sample**. A different implementation
ran, and neither is wrong. To be careful about what that is — agreement between two implementations
of one approximant, _not_ a bound on either one's accuracy.

**Example approximant choice.** One 8-second segment: `IMRPhenomD` **10.3 s** end to end vs
`IMRPhenomXPHM` **71.6 s** (LAL: ~0.7 s). That is JIT compilation, not generation, and it is the same
order as the 121 s/call recorded for precessing models before gwmock-signal #342 — so it corroborates
that earlier figure. The example ships `IMRPhenomD`; 72 s would be a poor first run. Caveat: these
are total CLI wall times, so they bundle JAX import with compilation.

## Two corrections to my own reasoning

I wrote a comment claiming the lazy alias mapping avoids importing JAX. **False** — JAX is a hard
dependency of `gwmock-pop` (`jax>=0.5.0`), so it is always imported. What `gwmock[jax]` actually
supplies is **ripplegw**. Comment corrected; the lazy mapping is still right, for the right reason.

I also first wrote the example with the `sandbox.zenodo.org` population URL copied from the sibling
examples — the ephemeral one. Sandbox records are purged, so a new example should not ship a
known-expiring input; it now reads the small in-repo population.

## Scope

**This selects a library, not a compute device.** `ripple` is JAX-based but runs through the same
per-event path as LAL. `simulate_segments` still has **zero callers in `src/`**, so nothing here
drives the batched on-device entry point — that needs the orchestration loop replaced
(`gwmock/device-call-site`), and is where GPU end-to-end actually becomes real.

A separate trap surfaced while testing and is worth flagging for the e2e work: a run whose segment
does not contain the population's events **reports success and writes pure zeros**. The shipped
example's `start-time` is aligned to its event with a comment saying why, and a test asserts non-zero
content.

## Summary by CodeRabbit

* **New Features**
  * Added configuration support for selecting waveform backends, including LALSimulation, PyCBC, Ripple, GWSIGNAL, and custom backends.
  * Added backend-specific constructor arguments and metadata recording for reproducible runs.
  * Added a Ripple waveform example with documentation and optional dependency guidance.

* **Documentation**
  * Documented backend selection, installation requirements, limitations, and first-use compilation timings.

* **Bug Fixes**
  * Expanded the spelling allowlist for waveform-related terminology.

### DIFF
--- a/.typos.toml
+++ b/.typos.toml
@@ -22,5 +22,8 @@ extend-ignore-re = [
 [default.extend-words]
 # numpy function name, not a misspelling of "arrange"
 arange = "arange"
+# a binary whose orbital plane precesses; read as "processing" otherwise
+precessing = "precessing"
+precession = "precession"
 aer = "aer"
 BA = "BA"
--- a/docs/user-guide/configuration.md
+++ b/docs/user-guide/configuration.md
@@ -297,7 +297,9 @@ orchestration:
 `signal.waveform-backend` selects which library generates the polarizations —
 `lal` (the default), `pycbc`, `ripple`, or `gwsignal`. It also accepts an entry
 point in the `gwmock.waveform` group or a `module:Class` reference, so a
-third-party backend can be plugged in the same way.
+third-party backend can be plugged in the same way. Such a backend is matched by
+its public surface — `available_approximants` and `generate_td_waveform` — and
+does not have to subclass gwmock-signal's `WaveformBackend`.
 
 Constructor arguments for that backend go under
 `signal.waveform-backend-arguments`. These are distinct from `signal.arguments`,
--- a/docs/user-guide/configuration.md
+++ b/docs/user-guide/configuration.md
@@ -292,6 +292,40 @@ orchestration:
                 channel: '{{ detectors }}:STRAIN'
 ```
 
+### Choosing the waveform library
+
+`signal.waveform-backend` selects which library generates the polarizations —
+`lal` (the default), `pycbc`, `ripple`, or `gwsignal`. It also accepts an entry
+point in the `gwmock.waveform` group or a `module:Class` reference, so a
+third-party backend can be plugged in the same way.
+
+Constructor arguments for that backend go under
+`signal.waveform-backend-arguments`. These are distinct from `signal.arguments`,
+which is passed to the _simulator_ rather than to the waveform backend:
+
+```yaml
+signal:
+    waveform-model: IMRPhenomD
+    waveform-backend: ripple
+    waveform-backend-arguments:
+        taper_fraction: 0.05 # ripple-specific
+```
+
+Two things to be aware of:
+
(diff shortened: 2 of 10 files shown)


isaac-cf-wong:
@coderabbit review

PR #289, opened 2026-07-31
test(e2e): run the example matrix through the CLI in its own suite

isaac-cf-wong:
## 📝 Summary

#287 declared an end-to-end matrix with **no runner** — the entries were checked only for
existence. This adds the runner.

- **`e2e` marker**, excluded from the default run via `-m "not integration and not e2e"`, plus its
  own CI job
- **`tests/e2e/overlay.py`** — declarative per-example overrides that shorten a run and repoint its
  inputs. The examples themselves are never edited, so they stay realistic while the suite finishes
  in about two minutes
- **`tests/e2e/test_examples_end_to_end.py`** — runs each entry through the CLI and verifies its
  output
- **`tests/e2e/test_overlay.py`** — overlay invariants, deliberately _not_ marked `e2e` so they run
  on every PR
- **`tests/e2e/test_reproducibility.py`** — runs each entry twice in separate processes and compares
  content hashes

### Output is checked against the run's own manifest

Each run records every file it wrote, with channels and a content hash. The output directory is
compared against that declaration **in both directions**, so a run that wrote one detector's frame
and dropped the rest fails — and so does one that writes a file its metadata doesn't mention.

That replaced a weaker first version which required only _some_ output, _some_ decodable file,
_some_ non-zero sample. Verified by deleting a frame and by planting a stray file; each breaks it.

### Reproducibility, in separate processes

All six runnable entries are reproducible. Each repetition runs in a **subprocess**, because
same-process agreement could come from a cached RNG, JAX's compilation cache, or LAL's
process-global registry — none of which survive a fresh interpreter. Since stored reference values
will be compared across CI runs on different machines, process independence is the property that
matters.

This is the prerequisite for comparing against stored references, which is deliberately **not** in
this PR.

## ✨ Type of Change

      expected)
- [x] 📝 Documentation update

No `src/` changes. Tests, CI config, and docs only.

- [x] **Unit Tests:** e2e **27 passed, 8 skipped** (~108 s); default suite **804 passed, 23
      skipped**, unchanged except for the new deselections.
- [x] **Manual Test:** deliberately misaligned the start time — both signal entries fail, which is
      what makes the signal assertion meaningful.
- [x] **Manual Test:** changed the seed where it's actually read — all nine output files differ, so
      the reproducibility probe discriminates.
- [x] **Manual Test:** confirmed hashes are non-null; a probe comparing `None` to `None` would pass
      vacuously.
- [x] **Manual Test:** all 21 pre-commit hooks pass on the committed tree.

## 🏗️ Checklist

Not applicable.

## Two tests that couldn't fail, and how they were caught

Both found by deliberately breaking things rather than by reading the code.

**The signal assertion.** It counted non-zero samples across every output file. With the start time
misaligned, `signal/bbh` failed but `quick_start` **passed** — it writes noise too, and the noise
alone satisfied the assertion. Now restricted to the configured signal output directory.

**The seed.** My first discrimination check changed the _global_ seed and saw **0 of 9** files
differ. Cause: the orchestrator copies the global seed into the noise arguments with `setdefault`,
so an example pinning its own (several pin `42`) keeps it. The overlay's global seed was a **no-op**
for those configs — meaning the runs were reproducible because the _examples_ pin a seed, not
because the test did. The overlay now writes the seed where it's read, and a test pins that.

## Known limits, stated rather than implied

**The gengli entry never runs.** It needs `gengli` — which is **not a declared extra of this
project**, so no sync flag installs it — and a local blip-glitch fixture, since the example reads
one from `sandbox.zenodo.org`, whose records are purged. It's labelled `**Not run**` in both the
matrix and the README, enforced by a test, because seven entries reading as seven exercised paths is
the most misleading thing a matrix can do.

**The overlay costs coverage.** Sampling frequency drops 4096 → 1024 Hz, moving the Nyquist
frequency, and spans are shortened, reducing segment counts. A defect appearing only at full rate or
across many segments is out of scope here.

**CI runs e2e on one Python and one OS.** These exercise orchestration, not anything
interpreter-specific; the normal matrix still covers portability. The job installs `--all-extras`
_and asserts the optional stack is importable_, because otherwise a missing extra makes the gated
entries skip and the job reports green while testing less than it claims.

## Review

An independent review reproduced both test counts and produced the manifest check, the subprocess
reproducibility runs, the extended preserved-key list, and the `**Not run**` labelling. It also
caught a false comment claiming `--all-extras` covers gengli. It could not reproduce the
seed-perturbation or misaligned-start experiments; those remain my measurements.

## Not in this PR

Stored reference values (tier 3). Also still open: `quick_start` and the other signal examples fetch
inputs from mutable `main` and purge-prone `sandbox.zenodo.org`. The overlay repoints them so the
_suite_ is hermetic, but those examples stay fragile, and pinned inputs are a prerequisite for
stored references meaning anything.

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive end-to-end coverage for example configurations, output generation, data validity, and expected signals.
  * Added reproducibility checks to confirm identical results across repeated runs.
  * Added validation for deterministic, isolated test configurations and required fixtures.

* **Chores**
  * Added automated end-to-end testing to continuous integration.
  * Updated test markers and documentation, including known unavailable scenarios.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -56,9 +56,13 @@ jobs:
     # them across seven interpreter/OS combinations would multiply that for no added signal,
     # since what they exercise is orchestration rather than anything interpreter-specific.
     #
-    # `--all-extras` is load-bearing. The matrix in tests/e2e/matrix.py has entries gated on
-    # ripplegw and gengli, and with a plain `uv sync` those would skip -- reporting green while
-    # testing nothing. One Python and one OS, but the whole optional stack.
+    # `--all-extras` is load-bearing for the ripplegw-gated entries: with a plain `uv sync` those
+    # would skip, and the job would report green while testing less than it claims. The guard step
+    # below turns that from a silent skip into a failure.
+    #
+    # It does *not* cover the gengli entry. gengli is not a declared extra of this project at all,
+    # so no sync flag can install it, and that entry is separately blocked on a local glitch
+    # fixture. It stays skipped here -- see NOT_HERMETIC in tests/e2e/overlay.py.
     e2e:
         runs-on: ubuntu-latest
         steps:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,3 +50,38 @@ jobs:
               uses: codecov/codecov-action@0fb7174895f61a3b6b78fc075e0cd60383518dac # v5
               with:
                   token: ${{ secrets.CODECOV_TOKEN }}
+
+    # End-to-end runs, kept out of the `test` matrix on purpose. They drive example configs
+    # through the real CLI and generate data, so they are minutes rather than seconds -- running
+    # them across seven interpreter/OS combinations would multiply that for no added signal,
+    # since what they exercise is orchestration rather than anything interpreter-specific.
+    #
+    # `--all-extras` is load-bearing. The matrix in tests/e2e/matrix.py has entries gated on
+    # ripplegw and gengli, and with a plain `uv sync` those would skip -- reporting green while
+    # testing nothing. One Python and one OS, but the whole optional stack.
+    e2e:
+        runs-on: ubuntu-latest
+        steps:
+            - uses: actions/checkout@d23441a48e516b6c34aea4fa41551a30e30af803 # v6
+
+            - name: Install uv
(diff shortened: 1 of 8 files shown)


isaac-cf-wong:
@coderabbitai review

PR #290, opened 2026-07-31
test(e2e): scope exact hashing to one numerical stack, and guard write mode against CI

isaac-cf-wong:
## 📝 Summary

The end-to-end matrix ran and checked itself structurally (#289), but nothing noticed if a
waveform, PSD or projection quietly changed. This adds the last tier: comparison against **stored
reference values**, so an unintended change in the data fails on the pull request that causes it.

Six JSON files under `tests/e2e/references/`, **28 KB total**, each recording per output file a
content hash plus `n`, `nonzero`, `argmax`, `peak`, `rms`, `mean`, and the versions of the packages
that could change the data.

### Exact hashing, but only where it is a fair question

The hash is asserted when the run happens on the **same numerical stack** the reference came from —
one changed sample fails, no tolerance, because a tolerance picked without a reason to justify it
silently accepts whatever falls inside it.

References therefore record a fingerprint (system, machine, Python). On a *different* fingerprint,
identical bits are not a reasonable thing to demand: the same numpy and LAL built against a
different BLAS can differ in the last bits. There the stored statistics are compared with a
relative tolerance and the test skips, naming both environments — so the weaker check is never
passed off as the strong one.

**The tolerance is declared unmeasured.** It sits below any physically meaningful change (a
different waveform, a rescaled PSD, a lost taper all move these by percent or more) and above float
noise. Nobody has measured where two platforms actually land, and the docstring says so.

### A failure is a prompt to look, not a verdict

A bare hash says something changed and nothing about what, so every failure becomes a bisect. The
statistics make it answerable. Real output, from perturbing the seed:

```
output/noise/E-ET1_SARD_STRAIN_NOISE-1577491296-16.gwf:
  argmax: 16381 -> 15972
  peak: 1.626e-22 -> 1.917e-22  (relative change 1.786e-01)
  rms:  4.238e-23 -> 5.442e-23  (relative change 2.842e-01)
no recorded package version changed, so a dependency bump does not explain this
```

18–35% shifts are obviously a different noise realisation rather than rounding, and that last line
is what separates a harmless upgrade from someone altering behaviour. When a change *is* harmless,
regenerate and review the diff:

```bash
GWMOCK_WRITE_E2E_REFERENCES=1 uv run pytest -m e2e --no-cov
```

Write mode **refuses to run when `CI` is set**. Otherwise a stray environment variable would let a
CI run overwrite the comparison and then report success.

## ✨ Type of Change

      expected)
- [x] 📝 Documentation update

No `src/` changes. Tests, generated reference data, one pre-commit exclusion, and docs.

- [x] **Unit Tests:** e2e **34 passed, 9 skipped** (132 s); default suite **804 passed, 23
      skipped**.
- [x] **Manual Test:** perturbed the seed — `peak`/`rms` move 18–35% and the report says no package
      version changed. Only the noise files move, correctly, since that entry's signal does not
      depend on the seed.
- [x] **Manual Test:** regenerating twice produces no diff, so the workflow is idempotent.
- [x] **Manual Test:** `CI=true GWMOCK_WRITE_E2E_REFERENCES=1` raises rather than rewriting.
- [x] **Manual Test:** a foreign fingerprint with agreeing statistics skips, naming both
      environments; a foreign fingerprint with a 2.9% peak change **fails** and reports the relative
      difference.

## 🏗️ Checklist

Not applicable.

## Restructure

The run-once-per-entry fixture moved to `conftest.py`, because a fixture defined in a test module is
invisible to other modules, and the shared helpers moved with it into `runner.py` so `conftest` does
not import a test module. No behaviour change; the structural and reproducibility tests use the same
helpers under public names.

References are excluded from prettier. It reindents the generator's two-space output to four, so
every regeneration would carry reformatting noise on top of the real change — defeating the point of
storing them as reviewable text. `check-json` still validates syntax.

## What the review changed

An independent review reproduced both counts and confirmed the references match on a second run. It
could not test a second machine, so portability stayed unresolved — but it caught that the
references recorded package versions while the test demanded identical bits from *everyone*, which
is what prompted the fingerprint scoping above. It also found the `None == None` hole (both sides
must now have a non-empty string hash), the CI write-mode hazard, and gave counterexamples showing
the statistics oversold themselves: a sign inversion leaves `peak` and `rms` untouched. Hence
`mean`, and a docstring that now says the remaining gaps are a limit of the explanation rather than
of the detection.

It **could not reproduce** the seed-perturbation experiment, nor test cross-machine or cross-BLAS
stability. Both remain my measurements.

## Still open

**Cross-machine portability is unverified.** The fingerprint mechanism means a mismatch now degrades
gracefully instead of failing spuriously, so this is contained rather than solved — but nobody has
confirmed that two machines with the same fingerprint really do produce identical bits. CI is the
first independent test.

The `gengli` matrix entry remains `**Not run**` and so has no reference. Also unchanged: the signal
examples still fetch inputs from mutable `main` and purge-prone `sandbox.zenodo.org`; the overlay
repoints them so the suite is hermetic, but those examples stay fragile.

## Summary by CodeRabbit

* **Tests**
  * Added end-to-end validation against stored output references for supported scenarios.
  * Added checks for missing, unexpected, or changed generated files, with clearer diagnostics.
  * Added environment-aware comparisons to distinguish expected runtime variation from unexpected changes.
  * Expanded coverage across default, quick-start, waveform, noise, and signal generation scenarios.

* **Documentation**
  * Documented reference comparisons, failure details, and the process for regenerating expected outputs.
  * Excluded generated reference files from automated formatting.


isaac-cf-wong:
@coderabbitai review

PR #291, opened 2026-07-31
fix(data): decide sample alignment with a spacing-derived absolute tolerance

isaac-cf-wong:
## 📝 Summary

Injection decided whether a chunk was grid-aligned with `np.isclose(offset, round(offset))`. That
tolerance is **relative**. A fractional offset can never exceed half a sample, so the test passes
unconditionally once `rtol * offset >= 0.5` — from about **50,000 samples, or 12 s into a segment at
4096 Hz**.

Past that point the interpolation branch was unreachable. Every chunk was snapped to the nearest
sample, displacing signals by up to **half a sample (122 µs at 4096 Hz)** behind a `logger.debug`.
The bundled examples use **4096-second segments** (up to 1.7e7 samples), so nearly every injection in
a realistic run was in that regime.

Both call sites were affected, and each carried its own copy of the check — `TimeSeries.inject`,
which interpolates _linearly_, and the module-level `inject`, which interpolates _cubically_. Both
now call one `is_aligned()`, so the rule has a single definition.

### The tolerance is derived, not chosen

`offset` comes from a GPS-scale subtraction, so even an exactly-aligned chunk carries cancellation
error. I measured it: **zero at power-of-two sample rates**, up to **4.0e-4 samples at 4000 Hz**
across GPS epochs 1e9–1.8e9 and rates 512–16384 Hz.

My first version used a fixed `1e-3`. Review found the regime that breaks: at **GPS 1e10 and 2000 Hz
the error reaches ~1.68e-3**, so a genuinely aligned chunk was classified misaligned and
interpolated, altering a smooth waveform by ~1.9e-3. That is the same defect as the original bug,
pointing the other way.

The error scales as `spacing(epoch) * sampling_frequency`, so a constant cannot be right everywhere.
The tolerance is now derived from `np.spacing` at the epoch times 4 — about an order of magnitude
above the measured 4.0e-4, kept as small as possible because **too loose is the harmful direction**.
Clamped below at 1e-9 for epochs near zero, and above at **1e-2** so it can never approach the half
sample that would make everything look aligned. Hitting that ceiling logs a warning, since it means
float64 cannot resolve alignment at that epoch and rate at all.

## ✨ Type of Change

      expected)
- [ ] 📝 Documentation update

No API is removed, but **this changes generated data** for configurations with segments longer than
~12 s at 4096 Hz — which is all the bundled examples. Signals that were silently snapped are now
interpolated to their true sub-sample position. Existing datasets will not reproduce bit-for-bit.

- [x] **Unit Tests:** 835 passed, 23 skipped (804 before, plus 31 here). e2e: 34 passed, 9 skipped.
- [x] **Manual Test:** red check with the old check restored — **49,999 / 50,001 / 200,000 fail; 10 /
      1,000 pass**. 49,999 fails because `round(49999.5) = 50000` and `rtol * 50000` is exactly 0.5,
      which is what makes the parametrisation straddle the real threshold rather than guess at it.
- [x] **Manual Test:** swept the cancellation error across epochs 1e9–1.8e9, rates 512–16384 Hz,
      offsets 0–1.7e7 to fix the floor at 4.0e-4.
- [x] **Manual Test:** confirmed the two-stage path does not interpolate twice — the linear pass
      leaves an integer offset, so the cubic branch is skipped. Now asserted.

## 🏗️ Checklist

Not applicable.

## Two mistakes of mine, since both are instructive

**The first test passed for the wrong reason.** It injected a flat top-hat and asserted the values
changed. Interpolating a constant returns the same constant in its interior, so a flat chunk cannot
reveal whether it was resampled or merely snapped. Replaced with a delta, whose energy visibly
spreads.

**A duplicate class name silently dropped 16 tests.** Adding
`TestAlignmentToleranceDerivation` I first named it `TestAlignmentTolerance`, shadowing the existing
class. 61 tests became 45 and everything stayed green. Caught from the count, not from a failure.

## What the end-to-end suite tells us, and what it cannot

The reference hashes are **unchanged**, and that is the expected result rather than a reassuring one:
the e2e overlay shortens segments to 4–16 s, keeping every offset below the threshold. So **the e2e
suite structurally could not have caught this bug** — exactly the limitation recorded in #289 as out
of scope, now demonstrated rather than hypothetical.

Worth considering separately: one matrix entry at full rate with a segment long enough to exceed 12 s
of offset would close that gap, at the cost of runtime.

## Not measured

I have **not** diffed a full 4096-second production segment before and after. The fix should change
such output, and that is the claim I have not verified directly — the independent review did not
verify it either. Everything above is either measured at test scale or reasoned from the arithmetic.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved time-series injection alignment checks for greater floating-point accuracy.
  * Correctly accepts genuinely sample-aligned injections while rejecting fractional-sample offsets.
  * Provides consistent alignment behavior across different reference times and sampling frequencies.
  * Warns when numerical precision is insufficient to reliably determine alignment.

### DIFF
--- a/src/gwmock/data/time_series/inject.py
+++ b/src/gwmock/data/time_series/inject.py
@@ -12,6 +12,36 @@ from scipy.interpolate import interp1d
 
 logger = logging.getLogger("gwmock")
 
+#: How close to a whole sample an offset must be to count as aligned, in samples.
+#:
+#: Absolute, not relative. ``np.isclose``'s default is relative, and since a fractional offset can
+#: never exceed 0.5 samples, a relative test passes unconditionally once ``rtol * offset >= 0.5`` --
+#: about 50,000 samples, or 12 s into a segment at 4096 Hz. Beyond that every chunk was snapped to
+#: the nearest sample and the interpolation branch was unreachable, silently displacing signals by
+#: up to half a sample.
+#:
+#: The value is measured rather than picked. ``offset`` comes from a GPS-scale subtraction, so an
+#: exactly-aligned chunk still carries cancellation error: zero for power-of-two sample rates, and
+#: up to 4.0e-4 samples at 4000 Hz across GPS epochs from 1e9 to 1.8e9. 1e-3 clears that worst case
+#: by 2.5x while still detecting a misalignment of a hundredth of a sample.
+#:
+#: Erring tight is the safer direction. Treating an aligned chunk as misaligned interpolates it onto
+#: essentially the same grid points, which is close to a no-op; treating a misaligned one as aligned
+#: is the displacement this constant exists to prevent.
+ALIGNMENT_TOLERANCE_SAMPLES = 1e-3
+
+
+def is_aligned(offset_samples: float) -> bool:
+    """Return whether *offset_samples* is a whole number of samples, to within tolerance.
+
+    Args:
+        offset_samples: Offset between two series' start times, in samples.
+
+    Returns:
+        Whether the offset counts as aligned. See :data:`ALIGNMENT_TOLERANCE_SAMPLES`.
+    """
+    return bool(abs(offset_samples - round(offset_samples)) <= ALIGNMENT_TOLERANCE_SAMPLES)
+
 
 def inject(timeseries: TimeSeries, other: TimeSeries, interpolate_if_offset: bool = True) -> TimeSeries:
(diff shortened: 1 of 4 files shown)


isaac-cf-wong:
@coderabbitai review

PR #292, opened 2026-07-31
feat(signal): convert batched device strain into gwmock chunks

isaac-cf-wong:
## 📝 Summary

First half of routing the CLI through the device path. `simulate_segments` has existed since #286 but
still has **zero callers in `src/`**; this adds the piece that makes wiring it possible, on its own so
it can be reviewed on its own.

`simulate_cbc_batch` returns strain of shape `(n_events, n_detectors, n_samples)` — **per-event and
unsuperposed**, with placement left to the caller. gwmock already has an assembler for exactly that
shape: `TimeSeries.inject_from_list` places chunks by absolute time, caches whatever overflows a
segment, and carries it into the next one.

So the device does generation and projection; gwmock keeps assembly.

### Why that split, rather than using `assemble_segments`

A measurement, not a compromise. Generation was **9.67e-3 s per second of data** against **3.66e-4 for
assembly** — so the expensive ~96% moves to the device while the part that owns spill-over,
resumability and memory bounds stays where it already works.

`gwmock_signal.assemble_segments` does the same job whole-span and stateless. Substituting it would
mean giving up gwmock's incremental, checkpointed model, which is what makes runs resumable and
memory-bounded. This closes `gwmock/assembly-ownership` as a decision with a number behind it rather
than a merge.

### Alignment is required, not hoped for

With an `output_grid`, each buffer starts exactly on a grid sample, so injection is an **integer-offset
add**. Without one, buffers begin at an arbitrary time and every chunk is resampled onto the segment —
"accurate only for heavily oversampled strain", in gwmock-signal's own words.

Three guards, each refusing something that would otherwise degrade quietly:

| Refused                      | Why it matters                                                                                                          |
| ---------------------------- | ----------------------------------------------------------------------------------------------------------------------- |
| Batch with no output grid    | Every chunk would be resampled                                                                                          |
| Buffer off the lattice       | Interpolated instead of placed                                                                                          |
| Detector order mismatch      | Channel _k_ becomes detector _k_ downstream — a permutation misattributes signals **without changing any shape**         |

`per_event_injections` keeps provenance per event, matching what the per-event loop records. Chunks
stay per-event precisely so that remains possible, which sidesteps
`gwmock/per-event-provenance-on-segments` for this slice rather than solving it.

## ✨ Type of Change

      expected)
- [ ] 📝 Documentation update

Purely additive. Nothing calls this yet, so no existing behaviour changes.

- [x] **Unit Tests:** 16 new; 820 passed, 23 skipped overall.
- [x] **Manual Test:** each guard verified load-bearing by deleting it — exactly one test fails in each
      case.
- [x] **Manual Test:** one test runs the **real** batched path, so the stub cannot drift from
      `BatchedDetectorStrain`'s actual fields and shapes.
- [x] **Manual Test:** converted chunks injected into a real segment; the second event's second
      detector lands **verbatim** at its grid index, which is the end-to-end proof that alignment gives
      an add rather than a resample.

## 🏗️ Checklist

Not applicable.

## On the test design

Mostly against a stub batch, for two reasons: the guards stay covered in an installation without the
`[jax]` extra, and the failure cases can be **constructed** rather than hoped for. Stub samples are
filled with `event + detector/10`, so a chunk that ends up in the wrong place is identifiable from its
contents rather than only from its length.

The real-device test exists because a stub can pass forever while the contract moves underneath it.

## Depends on #291

#291 reconciled gwmock's alignment tolerance with gwmock-signal's
`SamplingGrid.lattice_tolerance_samples`. That matters here specifically: this module places chunks on
gwmock-signal's lattice and they are injected using gwmock's tolerance, so a disagreement would mean a
buffer gwmock-signal calls on-lattice that gwmock calls misaligned. On `main` today the two differ by
a factor of two.

Merging this before #291 is not unsafe — the guard uses gwmock-signal's own `is_on_lattice` — but the
pair is only coherent with both.

## Not in this PR

Wiring `_simulate()`, and the config key to opt in. Also unresolved: whether CI has a GPU. The path is
correct under CPU JAX, but "the GPU path is exercised" needs a device, and those are different claims.
I would measure throughput rather than predict it.


isaac-cf-wong:
@coderabbitai review

isaac-cf-wong:
@coderabbitai review

isaac-cf-wong:
@coderabbitai review

isaac-cf-wong:
@coderabbitai review

PR #295, opened 2026-08-01
feat(cli): run a segment's events through the batched device path

isaac-cf-wong:
> **Stacked PR.** This targets `main`, so the diff includes the commits of #291 and #292, which are
> open and already reviewed. **The new work is the last two commits only** —
> `feat(cli): run a segment's events through the batched device path` and
> `fix(cli): honour waveform configuration in batched mode`. Merge #291 and #292 first.

## 📝 Summary

`simulate_segments` has existed since #286 with **zero callers**. This gives it one, and closes
`gwmock/device-call-site` — the blocker for GPU end-to-end since this work started.

A new signal key, `execution: per-event` (default) or `batched`, selects how a segment's events are
generated. `batched` calls `simulate_cbc_batch` once for the segment, converts the result to
per-event chunks (#292), and hands them to gwmock's existing assembler.

Default is unchanged, and an unknown value is **rejected** rather than falling back — a typo would
otherwise leave the run on the default path producing output that looks entirely normal.

### The batched path changes how events are _generated_, not how they are _assembled_

The branch sits inside `_simulate`, already called once per segment, and returns the same per-event
chunks the loop does. So spill-over into later segments and per-injection provenance keep working
unchanged. Buffers are generated against a `SamplingGrid` built from the segment's own start time, so
each begins on a sample of the segment lattice and injection is an integer-offset add.

### The two modes agree to 1.03e-12 of the peak

Same config, same library — so this isolates _batching_ rather than conflating a library change. Not
bit-identical, nor should it be: a batched `vmap` reassociates differently from a loop. A deliberate
1e-6 perturbation is caught, giving six orders of margin between real agreement and detectable
change.

## ✨ Type of Change

      expected)
- [ ] 📝 Documentation update

Default behaviour unchanged.

- [x] **Unit Tests:** 880 passed, 23 skipped; e2e 35 passed, 9 skipped. 16 new tests.
- [x] **Manual Test:** both modes end at the same `population_index`, and record **identical**
      injection provenance.
- [x] **Manual Test:** the agreement test is discriminating — a 1e-6 perturbation fails it.
- [x] **Manual Test:** `taper_fraction: 0.25` changes the buffer 3200 → 3600 samples, proving backend
      arguments reach the device.

## 🏗️ Checklist

Not applicable.

## Three things found by doing it

**The bundled catalogue broke batched mode.** It uses `distance`; the batched entry point wants
`luminosity_distance`. The per-event backends resolve that internally, so `execution: batched`
**failed on a config that worked** — switching mode would have changed whether a run happens at all.
Aliases now come from gwmock-signal's own tables, and a conflicting alias/canonical pair is refused
rather than silently resolved.

**Batched mode ignored the waveform configuration** — found by review, and the same silent-drop defect
as #288. The batched entry point takes a backend _instance_, which I never passed, so
`waveform-backend-arguments` were discarded and the run used ripple's defaults. Now passed through; a
non-ripple `waveform-backend` and `waveform-options` are refused rather than quietly served with
ripple.

**My first agreement assertion was wrong.** `np.allclose(..., atol=0)` applies a _per-sample_ relative
tolerance and fails wherever strain is near zero — 1e-34 against 1e-30 is a 1e-4 relative difference,
and most of a segment is near-zero. Comparing against the peak asks the question that matters. My own
rule about `atol=0` is right for equality checks and wrong for agreement ones.

## Claims I withdrew rather than defended

`_events_for_this_segment` originally claimed the two paths consume the catalogue identically. They do
not quite — the per-event loop has a second stop on the _generated_ strain's start time, which cannot
be evaluated before generating. Neither I nor the reviewer could construct a case where they differ
with the bundled waveforms, so the docstring now states which rule is shared, which is not, and what
reaching the difference would require.

## Known limits

**This has never run on a GPU.** Everything here is CPU JAX. "The path is correct" and "the GPU path
is exercised" are different claims, and only the first is supported.

**The alias table is a fourth copy** of a mapping gwmock-signal holds three times, one per backend.
The right home is a canonicalisation helper there; worth raising upstream rather than accumulating
copies.

**Duplicated boundary rule.** The event-consumption condition now exists in two places. The reviewer
recommended centralising it, which I would do as a follow-up rather than widen this branch.

## Summary by CodeRabbit

* **New Features**
  * Added configurable per-event and batched signal execution modes.
  * Added support for converting batched detector output into event-level strain chunks.
  * Added validation for execution settings, waveform backends, detector layouts, and output grids.
  * Preserved event provenance and checkpoint progress during batched processing.

* **Bug Fixes**
  * Prevented signal injection overflow handling from mutating or truncating caller-provided data.

* **Tests**
  * Added extensive coverage for both execution modes, validation, chunk conversion, provenance, and numerical equivalence.
  * Improved reference comparisons to tolerate expected floating-point differences.

### DIFF
--- a/src/gwmock/cli/adapter_orchestration.py
+++ b/src/gwmock/cli/adapter_orchestration.py
@@ -604,12 +604,14 @@ class AdapterOrchestrator(TimeSeriesMixin, Simulator):
             canonicalise_parameters,
         )
 
+        # Validated before any events are consumed. `population_index` is checkpointed state, so
+        # failing after advancing it would make a resumed run skip the events this call rejected.
+        waveform_backend = self._batched_waveform_backend()
+
         event_ids, events = self._events_for_this_segment()
         self._batch_injections = []
         if not events:
             return TimeSeriesList()
-
-        waveform_backend = self._batched_waveform_backend()
         parameters = canonicalise_parameters(
             {**self.waveform_arguments, **SignalAdapter.events_to_struct_of_arrays(events)}
         )
--- a/src/gwmock/cli/adapter_orchestration.py
+++ b/src/gwmock/cli/adapter_orchestration.py
@@ -532,16 +532,26 @@ class AdapterOrchestrator(TimeSeriesMixin, Simulator):
         events: list[dict[str, Any]] = []
         end_time_value = float(getattr(self.end_time, "value", self.end_time))
 
-        while self.population_index < len(self._population_events):
-            event_id = int(self.population_index)
-            parameters = dict(self._population_events[event_id])
+        # Read without advancing. `population_index` is checkpointed, and everything after this --
+        # transposing to a struct-of-arrays, canonicalising, generating -- can still fail. Advancing
+        # here would let a caller observe a consumed index after an exception, so the commit happens
+        # in `_commit_consumed_events` once generation has succeeded. The CLI's retry wrapper
+        # restores pre-batch state, so this is about direct library callers rather than the runner.
+        position = int(self.population_index)
+        while position < len(self._population_events):
+            parameters = dict(self._population_events[position])
             coa_time = parameters.get("coa_time")
             if coa_time is not None and float(coa_time) >= end_time_value:
                 break
-            event_ids.append(event_id)
+            event_ids.append(position)
             events.append(parameters)
-            self.population_index = cast(int, self.population_index) + 1
+            position += 1
         return event_ids, events
+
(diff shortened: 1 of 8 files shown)


isaac-cf-wong:
@coderabbitai review

isaac-cf-wong:
@coderabbitai review

isaac-cf-wong:
@coderabbitai review

PR #296, opened 2026-08-01
feat(signal): refuse settings the execution path would ignore

isaac-cf-wong:
> **Stacked.** Targets `main`, so the diff includes #292 and #295, both open. **The new work is the last two commits only.** Merge #292 and #295 first.

## 📝 Summary

Three defects in the batched execution path shared one shape: a setting was read from the configuration, stored on the orchestrator, and **never forwarded to the generator**. `waveform-backend-arguments` were discarded, `waveform-options` had nowhere to go, and `waveform-arguments` outside a fixed set were dropped. In each case the run completed and produced plausible output while disregarding what it had been told.

All three were instances of one default: **a setting nobody wired is ignored.** Fixing them individually leaves the next one waiting.

This inverts the default for the batched path. It declares which signal-config fields it honours, and anything else the user set is refused — so a configuration key added later fails loudly there until someone wires it. Pydantic's `model_fields_set` is what makes that workable: it distinguishes a field the user *wrote* from one left at its default, so the rule can be strict without refusing every configuration.

### Two further cases this surfaced

- **`signal.arguments`** configures the simulator's constructor, and the batched path calls gwmock-signal's batched entry point directly rather than going through the simulator — so these never apply.
- **`signal.parameters`** is read only by the stochastic path. For CBC sources it is ignored, and that predates the batched path entirely — it has been silently discarded on the default path all along.

Unrecognised keys are caught too, including nested ones. The config models allow extras, so a typo is *absent* rather than wrong, and nothing else reports it. `signal.output` is itself an `extra="allow"` model, so `output.<typo>` needs the same treatment even though `output` as a whole is honoured — otherwise the strict rule has an escape hatch one level down.

### Why the per-event path only warns

It is the default path, it has users, and `signal.parameters` has been ignored there for longer than the batched path has existed. Turning that into an error would break working configurations for a problem they did not introduce. The strict rule applies where the path is new and opt-in.

The check runs at the top of `_simulate`, before any event is consumed — `population_index` is checkpointed state, so failing after advancing it would make a resumed run skip the rejected events. It runs once per orchestrator rather than once per segment, so a warning is not repeated for every segment of a long run.

## ✨ Type of Change

- [ ] 💥 Breaking change
- [ ] 📝 Documentation update
- [ ] 🎨 Refactoring

The batched path is new and opt-in, so refusals there break nothing existing. Per-event behaviour is unchanged apart from two warnings.

- [x] **Unit Tests:** 902 passed, 23 skipped; e2e 35 passed, 9 skipped. 17 new tests.
- [x] **Manual Test:** all refusal cases verified — `waveform-options`, `arguments`, `parameters`, an unknown top-level key, an unknown nested `output` key, and a plain configuration that must still be accepted.
- [x] **Manual Test:** defaults are not treated as configured, which is what stops the rule refusing everything.
- [x] **Manual Test:** the per-event warnings are exercised through `_simulate` on a real orchestrator, not only by calling the helper — a test that calls the helper directly cannot tell a wired-up check from a dead one.
- [x] **Manual Test:** every new test confirmed failing against the unfixed code before being committed.

## 🏗️ Checklist

## The risk this carries

`_BATCHED_HONOURED_FIELDS` is derived by hand from what `_simulate_batched_segment` does. A field wrongly listed as honoured reproduces the original defect — a silently ignored setting — but now with a passing check beside it. That list is the part of this change most worth scrutinising.

## Not in this PR

`signal.parameters` being ignored for CBC on the **per-event** path is pre-existing and only warned about here. Fixing it properly means either wiring it through or removing it from the schema for non-stochastic sources, which is a behaviour change deserving its own PR.

## Summary by CodeRabbit

* **New Features**
  * Added configurable per-event and batched signal-generation modes.
  * Batched processing now preserves event chunks, provenance, and checkpoint safety.
  * Added validation with clear feedback for unsupported configuration options.
  * Added support for converting batched waveform output into individual event segments.

* **Bug Fixes**
  * Preserved metadata and channel identity when injected data extends beyond segment boundaries.
  * Prevented source data from being modified during boundary-overflow handling.

* **Documentation**
  * Improved reference-output comparisons and added signed peak diagnostics for portability.

### DIFF
--- a/src/gwmock/cli/adapter_orchestration.py
+++ b/src/gwmock/cli/adapter_orchestration.py
@@ -99,6 +99,7 @@ class AdapterOrchestrator(TimeSeriesMixin, Simulator):
         self._population_events = tuple(population_events)
         self._population_metadata = dict(population_metadata)
         self._source_type = source_type
+        self._configuration_checked = False
         self._source_detector_specs = tuple(source_detector_specs)
         self._signal_network = detector_network
         self._detector_resolution = detector_resolution
@@ -476,6 +477,8 @@ class AdapterOrchestrator(TimeSeriesMixin, Simulator):
 
     def _simulate(self) -> TimeSeriesList:
         """Generate signal chunks for the current segment from population events."""
+        self._require_configuration_supported()
+
         if not self._population_events and self._source_type == "sgwb":
             return self._simulate_stationary_signal_segment()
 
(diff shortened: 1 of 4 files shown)


isaac-cf-wong:
@coderabbitai review

PR #297, opened 2026-08-01
fix(data): report the signal content injection drops off the front of a segment

isaac-cf-wong:
## 📝 Summary

`TimeSeries.inject` crops a chunk to the target segment's span. Forward overflow is returned as a tail and carried into the next segment; **backward overflow is discarded**, because the segments it belongs to have already been written.

A compact-binary waveform starts before its `coa_time`, while the segment claiming an event is the one `coa_time` falls in. So a buffer reaches back past its own segment whenever `coa_time` lands within one buffer length of a boundary — an ordinary configuration, not an exotic one. The run completed, the frames looked normal, and a truncated inspiral is indistinguishable from a short one.

Measured at 1024 Hz with 16 s segments, IMRPhenomD:

| Case | Buffer | Dropped | Samples | Strain² energy |
|---|---|---|---|---|
| BBH 30+25 M☉ @20 Hz, `coa_time` 0.5 s past a boundary | 4.0 s | 3.1 s | 77.5% | **32.4%** |
| BNS 1.4+1.35 M☉ @20 Hz | 256 s | 230.4 s | 90.0% | **99.998%** |

Exposure scales as pre-`coa` buffer over segment length, so the shipped `duration: 4096` examples affect roughly 5.6% of BNS events rather than all of them. The 16 s test configs affect most.

**Nothing about the placement changes here.** The content is still dropped. A run now says which signal lost what — in seconds, samples, and the share of summed squares.

### Why energy rather than sample count

They differ by more than 2× in the BBH case (77.5% vs 32.4%), because the merger sits inside the retained window and the inspiral does not. Quoting the sample fraction would be the more misleading of the two.

It is **unweighted strain-squared energy, and explicitly not an SNR loss** — a matched-filter figure needs a detector PSD and frequency-domain weighting, neither of which `inject` has access to. The docstring says so.

### Why warn rather than refuse

The loss predates this change and refusing would break working configurations for a problem they did not introduce. The actual fix — claiming an event in the segment where its *waveform* starts — changes generated data and invalidates the e2e reference values, so it is tracked separately.

## ✨ Type of Change

- [ ] 💥 Breaking change
- [ ] 📝 Documentation update
- [ ] 🎨 Refactoring

Diagnostic only — no output changes, so the e2e reference values are untouched.

- [x] **Unit Tests:** 932 passed, 23 skipped; e2e 35 passed, 9 skipped. 16 new tests.
- [x] **Manual Test:** mutation-tested rather than merely run — removing the half-sample slack fails exactly the ULP test; reporting the sample fraction instead of energy fails exactly the three energy tests; moving the measurement back after the early return fails exactly the entirely-early test.
- [x] **Manual Test:** one test drives a real orchestrator with a `coa_time` 0.5 s past a boundary and confirms the warning fires on a genuine configuration.
- [x] **Manual Test:** an ordinary injection, and a tail landing exactly on a boundary, both stay silent — a warning on every segment would be ignored.

## 🏗️ Checklist

## Limits worth knowing

- **The half-sample threshold can hide up to one genuinely dropped sample.** That is deliberate: a sample nearer the boundary than half a sample period rounds *to* it. Measured margin — float64 spacing at a GPS epoch is 2048× under the threshold at 1024 Hz, 128× at 16384 Hz, and would need a 2.1 MHz rate to reach it.
- **The energy figures are this codebase's own measurement.** They are not anchored against an independent SNR tool.
- **A boundary-overlapping BNS chunk costs an O(samples × channels) pass.** Ordinary injections pay nothing — the sum is reached only once the condition trips.

## Summary by CodeRabbit

* **New Features**
  * Added reporting when waveform content falls before a segment boundary.
  * Reports discarded duration, sample count, and signal energy fraction where applicable.

* **Bug Fixes**
  * Improved handling of empty, silent, multi-channel, off-grid, and non-overlapping waveform chunks.
  * Preserved correct discarding of content that precedes the target segment.

* **Tests**
  * Added comprehensive coverage for boundary precision, warning behavior, missing metadata, and real waveform injection scenarios.

### DIFF
--- a/src/gwmock/data/time_series/inject.py
+++ b/src/gwmock/data/time_series/inject.py
@@ -3,7 +3,7 @@
 from __future__ import annotations
 
 import logging
-from typing import cast
+from typing import Any, cast
 
 import numpy as np
 from astropy.units import second  # pylint: disable=no-name-in-module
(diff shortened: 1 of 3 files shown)


isaac-cf-wong:
@coderabbitai review

PR #298, opened 2026-08-01
feat(examples): make the GPU path reachable, and gate batched execution end-to-end

isaac-cf-wong:
## 📝 Summary

The batched path could run on a GPU, but nothing a user installs or reads would get them there, and the e2e gate did not cover it. Four related gaps.

### `[jax]` said GPU and delivered CPU

The extra was commented as *"the on-device (GPU) signal path"*, but it resolves CPU-only wheels: `jax.default_backend()` returns `cpu` and `jax-cuda12-plugin` is absent. Installing it and setting `execution: batched` gives batched execution **on the CPU**, with nothing at runtime saying so.

The comment now states that plainly, and a new **`cuda` extra** adds `jax[cuda12]`. Kept separate rather than folded into `jax` because the CUDA wheels are Linux x86_64 only.

Verified on real hardware rather than by reading the lockfile: `uv sync --extra cuda` on an RTX 2080 Ti node installs `jax-cuda12-plugin` and `jax-cuda12-pjrt`, and JAX then reports `backend: gpu`, `devices: [CudaDevice(id=0)]`.

### No example selected the batched path

`grep -rn "execution" examples/` returned nothing, so the single config key that reaches the device path was undiscoverable from the examples. `signal/execution/batched` is the ripple example plus `execution: batched` — diff the two and the key is the only change.

Run end to end at 1024 Hz over 128 s: three ET frames, 4608 non-zero samples each, peak 8.4e-22.

### Adding `cuda` would have quietly slowed CI

The e2e job synced `--all-extras`, so a `cuda` extra would have pulled multi-gigabyte CUDA wheels onto a runner with no GPU on every run. It now names `sgwb` and `jax` explicitly. The cost is that a future extra must be added to that line deliberately, which the comment says.

### The batched path was not in the e2e matrix

Every execution path was gated except this one, so an orchestration regression here would have escaped the gate built for exactly that. The new entry reuses the ripple entry's overlay — same span, rate and population — so the two configs differ only by `execution` and their references are comparable.

- [x] **Unit Tests:** 940 passed, 23 skipped; e2e 41 passed (up from 35), 9 skipped.
- [x] **Manual Test:** the batched example run end to end on CPU, output checked for real signal rather than exit status — an early run wrote all zeros because a shortened window excluded the event, which is what a broken example also looks like.
- [x] **Manual Test:** `[cuda]` verified on an actual GPU (HTCondor, `stadius-nc-5`, RTX 2080 Ti, driver 580.159.04). **First time gwmock's device path has executed on a GPU** — every prior measurement was CPU JAX.
- [x] **Manual Test:** GPU/CPU agreement measured before adding the e2e entry, to confirm one entry can gate both.

## Can one reference gate both devices? Yes, measured

Raw-sample agreement was not sufficient to answer this, because the gate matches `n`, `nonzero` and `argmax` **exactly**:

| gated statistic | GPU vs CPU |
| --- | --- |
| `n`, `nonzero`, `argmax` | **exact match** on all three channels |
| `peak`, `rms`, `signed_peak` | worst divergence **8.99e-14** against a 1e-6 tolerance |

`argmax` landed on the same sample (83948 / 83949 / 83950), so the peak is not near a tie. The entry therefore passes on either backend with no tolerance change.

Stating the limitation honestly: a 10⁷ margin means the gate certifies *same physics*, not *same bits*. It cannot detect a GPU/CPU divergence, and is not intended to.

## Why the two devices differ at all

Investigated to a conclusion rather than waved at, because "floating point" is not an answer. Full write-up with all five refuted hypotheses is in the workspace at `_investigations/software/gwmock/2026-08-01-gpu-cpu-batched-discrepancy.md`.

The difference originates in **ripple's frequency-domain waveform generation**, not the FFT, the projection, or gwmock's own code. With byte-identical input, every downstream stage sits at eps level — production inverse-FFT kernel 1.8 eps, static projection 1.1, rotating projection 21.2 — against 1422 eps end-to-end. Ripple's FD phase differs by 3840 eps while its magnitude differs by 26, a 150:1 split.

Its character is a **global time shift of −2.32e-16 s**, which is 1.05 ulp of a seconds-scale internal quantity and explains **95.9%** of the phase-difference variance. That is 2.4e-13 of one sample: the same waveform, imperceptibly translated. Benign.

**Not established:** which backend is more accurate. CPU `irfft` is bit-identical to `numpy.fft.irfft` and GPU is not, but that plausibly reflects a shared implementation rather than accuracy; settling it needs a higher-precision reference that was not run.

## ✨ Type of Change

- [ ] 💥 Breaking change
- [x] 📝 Documentation update
- [ ] 🎨 Refactoring

No generated data changes, so existing e2e references are untouched; the only new reference is the new entry's.

## 🏗️ Checklist

## Also corrected

The ripple example still claimed gwmock *"does not yet drive ripple's batched on-device entry point from a configuration file."* It has since #295.

## Not in this PR

Per-PR GPU regression testing. GitHub-hosted GPU runners are larger runners, which are always billed and unavailable on this org's free plan; a self-hosted runner is possible but GitHub explicitly advises against self-hosted runners on public repositories, since a fork's pull request can execute arbitrary code on the machine. GPU verification stays manual via HTCondor.

## Summary by CodeRabbit

* **New Features**
  * Added a batched signal-execution example supporting CPU and GPU workflows with Ripple waveforms and ET detector output.
  * Added optional dependency configurations for CPU-based JAX and CUDA 12 GPU support.

* **Documentation**
  * Documented batched execution setup, requirements, behavior, and supported hardware.
  * Clarified Ripple backend usage and updated example guidance.

* **Tests**
  * Added end-to-end coverage and reference validation for batched signal generation.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -89,10 +89,18 @@ jobs:
             - name: Confirm the optional stack is present
               # Otherwise a missing extra turns the whole job into a silent no-op: every gated
               # entry skips and the job still passes.
+              #
+              # The required modules are read from `E2E_MATRIX` rather than listed here, so adding a
+              # matrix entry with a new `requires` cannot leave this check behind. `gengli` is the
+              # documented exception: it is not an extra of this project, and its entry is labelled
+              # "not run" -- see NOT_HERMETIC in tests/e2e/overlay.py.
               run: |
                   uv run --no-sync python -c "
                   import importlib.util as u, sys
-                  missing = [m for m in ('ripplegw', 'jax') if u.find_spec(m) is None]
+                  sys.path.append('tests')  # append, not insert: tests/signal would shadow stdlib signal
+                  from e2e.matrix import E2E_MATRIX
+                  needed = {m for e in E2E_MATRIX for m in e.requires} - {'gengli'}
+                  missing = sorted(m for m in needed if u.find_spec(m) is None)
                   sys.exit('e2e job is missing: ' + ', '.join(missing) if missing else 0)
                   "
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -56,9 +56,13 @@ jobs:
     # them across seven interpreter/OS combinations would multiply that for no added signal,
     # since what they exercise is orchestration rather than anything interpreter-specific.
     #
-    # `--all-extras` is load-bearing for the ripplegw-gated entries: with a plain `uv sync` those
+    # The `jax` extra is load-bearing for the ripplegw-gated entries: with a plain `uv sync` those
     # would skip, and the job would report green while testing less than it claims. The guard step
     # below turns that from a silent skip into a failure.
+    #
+    # Extras are named explicitly rather than `--all-extras`, because `cuda` pulls multi-gigabyte
+    # CUDA wheels that this runner has no GPU to use. The cost of naming them is that a new extra
+    # is not picked up here automatically -- add it to this line when one is introduced.
     #
     # It does *not* cover the gengli entry. gengli is not a declared extra of this project at all,
     # so no sync flag can install it, and that entry is separately blocked on a local glitch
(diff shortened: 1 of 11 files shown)


isaac-cf-wong:
@coderabbitai review

PR #299, opened 2026-08-01
test(e2e): pin which bundled ET presets the device path can run

isaac-cf-wong:
## 📝 Summary

Closes the tracker item asking whether the bundled ET presets actually work on the device path. They were run on a GPU (HTCondor, RTX 2080 Ti) rather than reasoned about, and the answer splits:

| presets | approximant | `execution: batched` |
| --- | --- | --- |
| `signal/bbh/<network>` ×4 | `IMRPhenomXPHM` | ✅ generates |
| `signal/bns/<network>` ×4 | `IMRPhenomPv2_NRTidalv2` | ❌ **refused** |

### The BNS presets cannot use the batched path

`IMRPhenomPv2_NRTidalv2` is precessing **and** tidal. Ripple implements precessing models (`IMRPhenomPv2`, `IMRPhenomXP`, `IMRPhenomXPHM`) and tidal models (`TaylorF2`, `IMRPhenomD_NRTidalv2`, `IMRPhenomXAS_NRTidalv3`), but not the combination — and the batched path only ever generates with ripple.

The failure is loud: refused before generating anything, with a message saying it is rejected rather than substituted. That is the guard from #286 doing its job; a silent substitution would have produced plausible-looking but wrong BNS waveforms. Nothing recorded that the *shipped configurations* were affected, which is what this adds.

### The BBH presets work, with a useful surprise

All four networks generate on device, including the 2L and triangle `CustomDetector` networks. Non-zero sample counts scale exactly with detector count (2×14580, 3×14580); peaks 0.89–1.04e-21.

`IMRPhenomXPHM` compiled **once in ~100 s and then served the other three networks from cache at ~0.1 s.** The concern recorded against this item was ~121 s *per call*; it is once per process, for a fixed approximant, backend options and traced shapes. Changing sample rate or segment length can force another compile.

## What is pinned

Two tests over the **shipped files**. `tests/signal/test_adapter_batch.py` already covers the refusal *mechanism* with a synthetic approximant; what it cannot catch is a preset drifting onto an approximant ripple lacks, or ripple gaining one it currently lacks. Both directions now fail loudly, so lifting the BNS limitation is a deliberate act.

They read ripple's module-level approximant tuple rather than constructing a `RippleBackend`, which raises `ImportError` without the `[jax]` extra — so they run in the default CI job instead of skipping. A third, gated test pins that the private tuple still equals `available_approximants()`.

## ✨ Type of Change

- [ ] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [x] 📝 Documentation update
- [x] 🎨 Refactoring (tests only)

No source changes: this is verification, tests and documentation.

- [x] **Unit Tests:** 943 passed, 23 skipped; e2e 41 passed, 9 skipped.
- [x] **Manual Test:** all eight ET presets driven through `execution: batched` on an RTX 2080 Ti.
- [x] **Manual Test:** the BNS presets confirmed to still generate on the per-event path (LAL, 3 detectors, 12288 non-zero, peak 8.16e-22) — the config headers claim this, so it is no longer only asserted.
- [x] **Manual Test:** both new tests confirmed to discriminate in both directions.

## 🏗️ Checklist

## Scope of the GPU check — what it does *not* establish

One event over a **128 s span at 1024 Hz**, with `execution: batched` injected. The presets themselves ship as per-event, and the real configurations are 4096 s segments at 4096 Hz over a day. This shows the approximant and the ET networks are usable on device; it does **not** show the shipped configurations fit in GPU memory. That remains untested.

## Review history worth knowing

The first version of these tests **passed vacuously** — they globbed `examples/<source_type>` rather than `examples/signal/<source_type>`, matched nothing, and asserted over an empty set. My red-checks had verified the comparison logic discriminated without verifying it was reading real files. The empty-result guards that now catch this are load-bearing.

Review also caught that constructing `RippleBackend()` would have failed the default CI job, and that an earlier draft of the README claimed the presets "run on device" when none of them set `execution` at all.

## Summary by CodeRabbit

- **Documentation**
  - Documented execution constraints for batched processing, including supported waveform types and BNS limitations.
  - Added benchmark guidance covering runtime conditions, GPU usage, caching, and shape-dependent recompilation.
  - Clarified why selected BNS presets use per-event execution.

- **Tests**
  - Added end-to-end validation for bundled BBH and BNS presets.
  - Added checks confirming supported waveform models match the backend’s available capabilities.
  - Improved preset discovery validation and error handling.

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -47,13 +47,24 @@ the installed JAX backend, not on the key: `gwmock[jax]` gives batched on the
 CPU, `gwmock[cuda]` gives a GPU. Check with
 `python -c "import jax; print(jax.devices())"`.
 
-**Not every preset can use it.** The batched path always generates with ripple,
-so a configuration naming an approximant ripple lacks is refused rather than
-served with a substitute. Verified on an RTX 2080 Ti on 2026-08-01: the four
-`signal/bbh/<network>` presets run on device (`IMRPhenomXPHM`, ~100 s to compile
-once, then cached across networks), while the four `signal/bns/<network>`
-presets cannot — `IMRPhenomPv2_NRTidalv2` is precessing _and_ tidal, a
-combination ripple does not implement.
+**Not every preset can use it.** None of the presets set `execution` — they all
+default to per-event — but if you add `execution: batched` to one, whether it
+works depends on its approximant, because the batched path only ever generates
+with ripple. A configuration naming an approximant ripple lacks is refused
+rather than served with a substitute.
+
+Measured on an RTX 2080 Ti on 2026-08-01, by adding the key to each preset: the
+four `signal/bbh/<network>` presets generated (`IMRPhenomXPHM`), while the four
+`signal/bns/<network>` presets were refused — `IMRPhenomPv2_NRTidalv2` is
+precessing _and_ tidal, a combination ripple does not implement. That run used a
+shortened 128 s span at 1024 Hz; it shows the approximant and the ET detector
+networks are usable on device, **not** that the shipped 4096 s, 4096 Hz, one-day
+configurations fit in GPU memory.
+
+`IMRPhenomXPHM` compiled once in ~100 s and then served the other three networks
+from cache. That caching holds for a fixed approximant, backend options and
+traced shapes; changing the sample rate or segment length can force another
+compile.
 
 **By detector network** — every `<network>` above is one of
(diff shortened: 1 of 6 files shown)


isaac-cf-wong:
@coderabbitai review

PR #302, opened 2026-08-02
feat(cli): make continuous waves reachable from a configuration file

isaac-cf-wong:
## 📝 Summary

Wires the `cw` source type through the orchestrator, so a pulsar catalogue plus
an ephemeris in a YAML file produces frames. Requires gwmock-signal at
`589da1c` (pinned in `[tool.uv.sources]`, development-only; to be replaced by a
version floor once that package is released).

Neither existing route fits a continuous wave. The stationary branch fires only
when there are *no* population events, since a stochastic background has no
sources. The per-event loop consumes events and expects a `coa_time`. A
continuous wave is stationary *and* has sources — a catalogue of pulsars, all of
which are present in every segment, none of which is ever consumed. So `_simulate`
gains a third branch, `_simulate_continuous_wave_segment`, which walks the whole
catalogue each segment, sums by feeding the running total back as the next
source's background, and deliberately never advances `population_index`.

The property the design protects is phase coherence across segment boundaries.
`reference_time_ssb` is a run-level constant reaching the backend through
`signal.arguments`; derived per segment instead, the phase would reset at every
boundary while every individual frame still looked correct. A test compares two
64 s segments against one 128 s run and holds them to 2.5e-12 of peak.

`sort_by` defaults to `coa_time`, which pulsars lack, so ordering is skipped when
the key is absent from every event, left at its default, *and* the source type is
`cw`. Absent from only some events still raises, as does an explicitly requested
key that is missing. The source-type condition matters: without it a compact
binary catalogue that lost the column — a header typo, a dropped field — would
stop raising and run unsorted, and since the per-event loop never breaks on a
missing `coa_time`, every event would land in the first segment. A full-looking
run with every coalescence time wrong.

`execution: batched` is refused with a message naming the reason. It was already
rejected before this change, because a continuous wave always sets
`signal.arguments`, but the error blamed the arguments rather than the source type.

### Known limitation

`signal_index.yaml` keys frames by signal id, so a source appearing in every
segment retains only its last frame. This is the pre-existing per-event
provenance behaviour, which continuous waves take from occasional to universal;
it is recorded at the code site. The per-batch metadata remains complete, so only
the `find-signal --id` lookup is lossy.

## ✨ Type of Change

- [x] **Unit Tests:** `uv run pytest` — 949 passed, 23 skipped
- [x] **Manual Test:** ran the shipped example end to end (shortened to a single
      64 s segment at 256 Hz), which wrote three frames for the ET triangle.

Six new tests in `tests/cli/test_continuous_wave_orchestration.py`. The two that
guard subtle failures were checked by mutation: the summation test fails against
an implementation that overwrites rather than accumulates, and the ordering test
fails when the source-type gate is removed. The earlier summation test did not —
it passed against an overwrite, catching the bug only because the loader happens
to return CSV rows in reverse order, so it now compares against the two sources
generated separately and added.

Not covered: an external reference for the absolute phase. The coherence test
would not catch a reference epoch that is wrong by a constant, since both sides
shift together. Only a LAL comparison settles that, and it is blocked on the
SWIG binding.

## 🏗️ Checklist

## Summary by CodeRabbit

* **New Features**
  * Added continuous-wave pulsar simulation support, including ET Sardinia detector configuration.
  * Continuous-wave signals now remain phase-coherent across segmented simulations.
  * All catalogue sources are included in every simulation segment, with source provenance recorded.
  * Added documentation and an example configuration for continuous-wave simulations.

* **Bug Fixes**
  * Multi-source continuous-wave output is now correctly combined into a single strain result.

* **Limitations**
  * Batched execution is not supported for continuous-wave simulations.

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -32,7 +32,7 @@ end.
 | Stochastic background                     | `signal/sgwb/<network>`                        |
 | Signals from a different waveform library | `signal/waveform_backend/ripple`               |
 | Batched generation (the GPU-capable path) | `signal/execution/batched`                     |
-| Continuous waves from pulsars             | `signal/cw/<network>`                          |
+| Continuous waves from pulsars             | `signal/cw/et_triangle_sardinia`               |
 | A blank starting point                    | `default_config`                               |
 
 **By waveform library** — `signal.waveform-backend` chooses which library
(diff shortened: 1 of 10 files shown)


isaac-cf-wong:
@coderabbitai full review

PR #305, opened 2026-08-03
chore(deps): consume gwmock-signal 0.13.0 and take over the ripple pin

isaac-cf-wong:
## 📝 Summary

gwmock-signal 0.13.0 is the first release carrying the continuous-wave simulator,
the device projection path, and the guard that refuses a non-finite geocentre
signal. The git pin on gwmock-signal comes out; the floors move to `>=0.13.0` in
`dependencies` and the `sgwb` / `jax` / `cuda` extras.

### The ripple pin has to move here, in the same change

Until now this file pinned *gwmock-signal* to a git revision, and that was
sufficient on its own: gwmock-signal carries its own `[tool.uv.sources]` pin of
ripplegw to the geocentre fix (GW-JAX-Team/ripple#141 — merged upstream as `ee148b6`, but not yet in a release), and uv
honours a git dependency's sources.

A wheel does not carry `[tool.uv.sources]`, so consuming the release breaks that
inheritance. Measured, not assumed — a scratch project depending only on
`gwmock-signal[jax]>=0.13.0` locks **ripplegw 0.3.0**, the version without the
fix. Without a pin here, continuous waves would not work in this repository's own
development or CI. They would fail loudly rather than silently, since 0.13.0
refuses NaN output, but they would not work.

### `[tool.uv.sources]` alone was not enough

uv applies a source to a package the project *declares*, and ripplegw reaches
gwmock only transitively through `gwmock-signal[jax]`. Adding the sources entry
changed nothing — the lock still resolved 0.3.0 from the registry — until
`ripplegw>=0.3` was declared in the `dev` group. A dev group is the right home:
development-only, so a git dependency cannot reach the published wheel.

Caught by re-reading `uv.lock` rather than trusting the edit. A pin that silently
does nothing looks identical to one that works, until continuous waves produce
NaN.

### A comment that stopped being true

The note claiming "one pin is enough" was correct only while gwmock-signal was a
git dependency. Rewritten, with the measurement, rather than left to mislead.

## ✨ Type of Change

- [x] **Unit Tests:** `uv run pytest` — 965 passed, 23 skipped
- [x] **Manual Test:** built the wheel with `uv build --no-sources` and read its
      metadata; re-resolved the lock; ran the continuous-wave suites against the
      released gwmock-signal.

| check | result |
|---|---|
| lock | gwmock-signal 0.13.0 (PyPI), ripplegw 0.3.1.dev3 (fork) |
| wheel metadata | `gwmock-signal>=0.13.0`, no git URL, ripplegw absent |
| continuous wave | 20 tests pass against the released gwmock-signal |

## 🏗️ Checklist

## Both pins come out together

#141 is merged upstream; what is missing is a ripplegw *release*. When one
appears: delete `[tool.uv.sources]`, raise the `ripplegw>=0.3` floor in the `jax`
and `cuda` extras to that release, do the same in gwmock-signal, and re-lock. The
four steps are written into `pyproject.toml`, with the reason they have to be done
as a unit — leaving either half alone resolves ripplegw 0.3.0 again.

The pinned revision `400afb4` is an ancestor of upstream `main`, so it stays
fetchable even after the source branch is deleted. That was worth checking: a pin
to a PR-head commit that only lived on a deleted branch would break every lock and
CI run. Until then continuous waves need a checkout, not a
PyPI install — and a PyPI install says so rather than writing NaN.

Not addressed here, tracked as `gwmock/cw-tests-never-run-in-ci`: gwmock's own
continuous-wave orchestration tests gate on a locally cached ephemeris, so they
skip in CI, and the CW e2e entry is marked not-run for the same reason. The fix
ports from gwmock-signal #364.

## Summary by CodeRabbit

* **Chores**
  * Updated minimum compatibility requirements for the signal-processing package and optional integrations.
  * Added RippleGW to the development toolchain.
  * Updated development dependency sourcing for RippleGW.

### DIFF
--- a/examples/signal/cw/et_triangle_sardinia/config.yaml
+++ b/examples/signal/cw/et_triangle_sardinia/config.yaml
@@ -17,10 +17,14 @@
 # whatever it is handed, which restarts the phase at every segment boundary -- producing data where
 # each segment looks perfectly correct and a coherent search over the run finds nothing.
 #
-# Requires the `jax` extra, and a gwmock built against a gwmock-signal that has the continuous-wave
-# simulator. Until that is on PyPI, `pip install 'gwmock[jax]'` will resolve a released
-# gwmock-signal without it and this example will fail to find the `cw` source type; install from a
-# checkout instead, where `[tool.uv.sources]` pins the right revision:
+# Requires the `jax` extra:  pip install 'gwmock[jax]'
+#
+# That is enough to resolve the continuous-wave source type -- gwmock-signal 0.13.0 registers it --
+# but not enough to *run* it. The simulator generates at the geocentre so one projection route
+# serves every source type, and released ripplegw returns NaN for every sample there
+# (`arccos(lz / rd)` is 0/0 at the Earth centre). It is refused rather than written, so a run fails
+# with a message naming the cause. Until GW-JAX-Team/ripple#141 reaches a ripplegw release, this
+# example needs a checkout of gwmock, where the pin in pyproject.toml supplies the fixed revision:
 #
 #     uv sync --extra jax
 #
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -22,7 +22,7 @@ classifiers = [
 requires-python = ">=3.12"
 dynamic = ["version"]
 dependencies = [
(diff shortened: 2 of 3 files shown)


isaac-cf-wong:
@coderabbitai full review

PR #306, opened 2026-08-03
chore(deps): lock file maintenance, and record astropy in the e2e references

isaac-cf-wong:
## 📝 Summary

Renovate's lock-file maintenance (#304) failed `e2e`:

```
AssertionError: 'noise/uncorrelated_gaussian/quick_start' differs from its reference
by more than floating-point drift.
  output/signal/E-ET3_SARD_STRAIN_BBH-1577491296-16.gwf:
    peak 7.612431335270193e-22 -> 7.612419388814761e-22 (relative 1.569e-06)
```

Investigated: the reference test doing its job, not a regression. This PR carries
the lock bump together with the references it moves, since the bump cannot land
without them — **it supersedes #304**, which can be closed.

## Cause, measured

`astropy-iers-data` 0.2026.7.27 → 0.2026.8.3 stepped UT1−UTC at the test epoch by
**2.740 ms**:

| | 0.2026.7.27 | 0.2026.8.3 |
|---|---|---|
| UT1−UTC | −0.044955 s | −0.047695 s |
| GMST at GPS 1577491296 | 1.7630853995385887 | 1.7630851997783947 |

ΔGMST = **1.9976e-07 rad**, against `2.740e-3 s × 7.2921e-5 rad/s` = **1.9980e-07**
analytically — agreeing to 0.02%.

**What is actually moving is not a revised measurement.** The examples run at
2030-01-01, about **885 days beyond the end of the packaged IERS table** (MJD 62502
against 61617). There is no data there; Astropy clamps UT1−UTC to the table's final
value. So each weekly release moves the *clamped edge* as the table's end date
advances. The arithmetic above holds either way, but the distinction matters: the
epoch rides the table edge, which makes this churn **structural** rather than
incidental. An epoch inside the table would move only when a finalised value is
revised.

gwmock-signal's projection takes sidereal time from Astropy deliberately, so it
uses real IERS UT1−UTC instead of assuming zero. The rotation therefore reaches
the antenna pattern and the strain.

All three detectors moved — ET1 5.7e-07, ET2 5.1e-07, ET3 1.57e-06 — and only ET3
crossed the 1e-06 gate. The 7.85× ratio between the GMST rotation and the peak
shift is **not** simply a sub-unity antenna response: ET3's factors near the peak
are (0.554, −0.214), norm 0.594, whose inverse is only 1.68. The remainder comes
from the plus/cross combination and a 2.27 ns change in the geocentre delay.

### Verified as benign, not assumed

Holding gwmock, gwmock-signal 0.13.0, ripplegw, Astropy, NumPy, SciPy, JAX and
LALSuite fixed and reverting **only** `astropy-iers-data` reproduces every old
`quick_start` statistic exactly, including ET3's peak. So #305 contributes none of
the shift. The shape agrees too: noise and SGWB content hashes unchanged, sample
counts and peak positions unchanged, and RMS deltas repeating detector-wise across
independent signal examples.

## The actual defect was the diagnostic

`_RECORDED_PACKAGES` listed gwmock\*, numpy, scipy, lalsuite, ripplegw and jax —
**no astropy**. So the report said:

> no recorded package version changed, so a dependency bump does not explain this

That points the next reader at a code regression that does not exist, for what is
the *most frequent* reason this comparison moves: the IERS table is republished
weekly. Both packages are now recorded, so the next bump explains itself.

All seven references are regenerated. The signal-bearing ones shifted slightly;
every one now carries the astropy versions.

## ✨ Type of Change

- [x] **Unit Tests:** `uv run pytest` — 965 passed, 23 skipped
- [x] **Manual Test:** `uv run pytest -m e2e --no-cov` — **41 passed, 15 skipped**
      (was 1 failed, 40 passed). Also installed both `astropy-iers-data` versions
      in scratch venvs and compared GMST directly, which is where the table above
      comes from.

## 🏗️ Checklist

## Reproducibility: a real hole, now closed for the tests and documented for users

**`--isolate` does not pin Earth-orientation data.** It pins package versions, and
`astropy-iers-data` supplies the table *bundled* with a release — but
`iers.conf.auto_download` defaults to `True` with an `auto_max_age` of 30 days.
Once the pinned release is older than that, Astropy fetches the current table from
the IERS server, and no recorded package version captures which one it got. A run
reproduced within a month of its dependencies' release matches; a year later,
sidereal time can differ.

Two responses, because the test suite and a user's run need different things:

- **The e2e runner now pins it.** The CLI is launched with auto-download disabled
  and a redirected Astropy cache, so references have inputs determined by what the
  references record. No reference value moved, confirming a fresh package was
  already using the bundled table — this makes it guaranteed rather than
  incidental. Asserted by reading the flag back from a child launched with the
  runner's environment, and mutation-checked.
- **The docs now say so.** `reproducibility.md`, `design.md` and `walkthrough.md`
  claimed bit-for-bit reproduction without this exception; they carry the caveat
  and the `auto_download = False` workaround.

## The churn is a standing decision, not fixed here

Tracked internally as `gwmock/iers-reference-churn` (a workspace tracker, not a
GitHub issue). Because the epoch rides the table edge, a weekly IERS release will
keep moving these references. Four options, and the last is new:

1. regenerate on each bump — honest, but a recurring review where a real
   regression could hide in the noise;
2. pin `astropy-iers-data` for e2e — stable references, ageing Earth-orientation
   data;
3. widen the drift tolerance past ~2e-06 — weakens every comparison to hide one
   known cause, and would stop it catching a projection regression of the same
   size;
4. **run the examples at an epoch inside the IERS table** — removes the cause
   rather than trading detection against noise. Changes every reference once.

The tolerance stays at 1e-06 in this PR, with the docstring corrected to say that
one legitimate drift now exceeds it.


isaac-cf-wong:
@coderabbitai full review

PR #308, opened 2026-08-03
test: run the continuous-wave tests in CI instead of skipping them

isaac-cf-wong:
## 📝 Summary

The continuous-wave tests ran **nowhere** in CI, and nothing reported that. Two
independent gates, with the suite green through both:

- the `test` job installs `--group dev` only, so every `importorskip("ripplegw")`
  skipped — ripple lives in the `jax`/`cuda` extras;
- `tests/cli/test_continuous_wave_orchestration.py` additionally skipped unless the
  LALPulsar ephemeris tables were already cached, which on a fresh runner they
  never are, because ripple fetches them on demand;
- the e2e entry sat in `NOT_HERMETIC`, labelled **not run**, for the same reason.

So the `_simulate` branch added in #302, and the device projection behind it, were
covered only on machines that happened to have run continuous waves before.

## What changed

A single-cell `test-jax` job installs the `jax` extra and runs the unit suite. The
ephemeris is cached with the identifier derived from `tests/data/ephemeris.sha256`,
and verified against those hashes before the tests run — they are physics inputs,
taken from lalsuite `master` and keyed by name alone, so a changed upstream table
or a corrupt cached copy would otherwise be used silently. A failed fetch fails the
job; skipping is what produced this situation.

**The same steps are on the `e2e` job**, because jobs do not share a filesystem and
the continuous-wave entry runs there.

**The job asserts its outcome, not just its prerequisites.** Installing the extra
and verifying the tables removes *reasons* to skip; nothing would notice a
re-added `skipif`, a renamed module or a changed marker quietly emptying the job —
the exact failure this fixes, reintroduced by one line. It runs the two
continuous-wave modules by name, reads back the JUnit report, and fails on zero
collected or anything skipped.

The e2e entry is out of `NOT_HERMETIC` with an overlay, kept multi-segment since
one segment cannot distinguish a never-consumed population from the per-event path,
and added to `CONTAINS_SIGNAL` because an always-on source fills every span.

## Two bugs the skip gate was hiding

- **The module hardcoded `~/.cache/ripplegw/ephemeris`** — the *Linux* default.
  Ripple uses `~/Library/Caches` on macOS and honours `RIPPLEGW_CACHE_DIR` above
  both, so it would have skipped on macOS even where the tables were present. It
  now asks ripple.
- **Running the e2e entry failed immediately**: `FileNotFoundError:
  ../../cw_population.csv`. The example points at its catalogue with a path
  relative to the examples tree, which does not survive being copied into a test
  working directory. Repointed like every other entry's population.

## ✨ Type of Change

- [x] **Unit Tests:** `uv run pytest` — 967 passed, 23 skipped
- [x] **Manual Test:** `uv run pytest -m e2e --no-cov` — **47 passed, 9 skipped**,
      up from 41 passed / 15 skipped. Also exercised `verify_ephemeris.py` against
      the real tables, a corrupted copy, a missing table, an empty manifest and a
      malformed line; and mutation-checked the ran-for-real guard by injecting a
      `pytest.mark.skip`, which fails it.

The new reference was reviewed rather than accepted: six outputs for three
detectors across two segments, 2048 samples each at 8 s and 256 Hz, and **every**
sample non-zero — the signature of an always-on source, where a transient would be
mostly zeros. Peaks differ per detector and shift slightly between consecutive
segments, as a rotating antenna pattern should.

## 🏗️ Checklist

## Known limits, named rather than buried

- **macOS continuous-wave coverage is still absent.** The job is one cell, so the
  platform-dependent cache lookup this PR *fixes* is never exercised on macOS.
  Widening to the full six cells costs six ephemeris fetches and six JAX installs
  per run — worth deciding rather than assuming.
- **`verify_ephemeris.py` and the test module read ripple's private `_cache_dir`.**
  Preferred over reimplementing the platform logic, and it fails loudly if it
  moves, but it is a private API in a package currently pinned to an unreleased
  revision. Re-check when that pin comes out.
- **The new reference has two moving inputs**: `astropy-iers-data`, whose weekly
  release already exceeds the drift tolerance (`gwmock/iers-reference-churn`), and
  the DE405 tables. Updating the hashes requires regenerating the reference in the
  same commit; nothing enforces that, so it is documented in the manifest.

## Summary by CodeRabbit

* **Tests**
  * Added continuous-wave end-to-end coverage for the Sardinia Einstein Telescope triangle scenario.
  * Continuous-wave tests now run with validated ephemeris data instead of being skipped when local tables are unavailable.
  * Added reference outputs and a lightweight local fixture for reliable, reproducible testing.
* **Chores**
  * Improved automated validation, caching, downloading, and integrity checks for ephemeris data across test environments.
  * Added JAX-specific test coverage and reporting.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -67,6 +67,86 @@ jobs:
     # It does *not* cover the gengli entry. gengli is not a declared extra of this project at all,
     # so no sync flag can install it, and that entry is separately blocked on a local glitch
     # fixture. It stays skipped here -- see NOT_HERMETIC in tests/e2e/overlay.py.
+    test-jax:
+        # The ripple/JAX-gated unit tests, which the `test` job cannot reach: it installs the dev
+        # group only, so every `importorskip("ripplegw")` there skips. That left the
+        # continuous-wave orchestration tests -- and the device-path tests -- running nowhere in
+        # CI while the suite reported green.
+        #
+        # One cell rather than the six of `test`: this exists to run the gated tests at all, and
+        # the code it covers is the same on every interpreter. Widen it if a version-specific
+        # failure ever appears.
+        runs-on: ubuntu-latest
+        # Pins ripple's ephemeris cache to one path. Its default is platform-dependent, so caching
+        # a hard-coded path would silently miss; RIPPLEGW_CACHE_DIR is a *base*, and ripple appends
+        # `ripplegw/ephemeris` to it, which is why the cached path is the base rather than the leaf.
+        env:
+            RIPPLEGW_CACHE_DIR: ${{ github.workspace }}/.ephemeris-cache
+            EPHEMERIS_FILES: earth00-40-DE405.dat.gz sun00-40-DE405.dat.gz
+        steps:
+            - uses: actions/checkout@d23441a48e516b6c34aea4fa41551a30e30af803 # v6
+              with:
+                  persist-credentials: false
+
+            - name: Install uv
+              uses: astral-sh/setup-uv@37802adc94f370d6bfd71619e3f0bf239e1f3b78 # v7
+              with:
+                  python-version: '3.12'
+                  enable-cache: 'true'
+
+            - name: Install dependencies with the JAX extra
+              run: uv sync --group dev --extra jax
+
+            - name: Confirm ripple is importable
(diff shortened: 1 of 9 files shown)


isaac-cf-wong:
@coderabbitai full review

PR #309, opened 2026-08-04
test(e2e): run the examples at an epoch inside the IERS table

isaac-cf-wong:
## 📝 Summary

The e2e reference comparison failed roughly **weekly**, on something that was not a
defect. This removes the cause.

The shipped examples run in 2030 to match the Einstein Telescope era, which is about
**885 days beyond the end** of the Earth-orientation table Astropy ships. There is no
data there, so Astropy **clamps** UT1−UTC to the table's final value — and every
weekly `astropy-iers-data` release extends that end date and moves the clamped value
with it. Measured: a 2.740 ms step, 1.9976e-07 rad of GMST, **1.569e-06 of peak** on
one detector, above this suite's 1e-06 drift tolerance.

Because the epoch rode the table edge, the churn was **structural**, not incidental.

## The evidence

Measured across the 0.2026.7.27 and 0.2026.8.3 releases, on GMST — the quantity the
projection actually consumes:

| epoch | 0.2026.7.27 | 0.2026.8.3 |
|---|---|---|
| 2030 (old) | 1.763085399538589 | 1.763085**199778395** |
| 2025 (new) | 1.760886943642783 | 1.760886943642783 |

Bit-identical at the new epoch.

## Why this option and not the others

Three alternatives were considered and are worse:

1. **Regenerate on each bump** — honest, but a recurring review where a real
   regression hides in the noise.
2. **Pin `astropy-iers-data` for the suite** — stable references, at the cost of
   testing against ageing Earth-orientation data.
3. **Widen the tolerance past ~2e-06** — would stop the check catching a projection
   regression of the same size, which is exactly what it exists for.

Closes the standing decision tracked as `gwmock/iers-reference-churn`.

## Why 2025-01-01 specifically

Finalised IERS data ended 2026-07-23 when this was chosen, so the epoch keeps about
**eighteen months of margin** before the prediction boundary — predictions get
revised, finalised values effectively do not. Verified at 2024-01-01, 2025-01-01 and
2026-01-01: all bit-identical across the two releases, while 2030 moved.

## What moved with it

- **The population fixture is now test-owned.** Its coalescence time has to sit at
  this epoch and the examples deliberately stay in 2030. Identical to the example's
  copy apart from that time.
- **The continuous-wave entry's `reference_time_ssb`.** It could have stayed in 2030,
  since it is only the epoch the spin parameters refer to, but the run would then
  describe a catalogue referenced to a time it never covers.
- **All eight references**, regenerated.

## ✨ Type of Change

- [x] **Unit Tests:** `uv run pytest` — 967 passed, 23 skipped
- [x] **Manual Test:** `uv run pytest -m e2e --no-cov` — 47 passed, 9 skipped. The
      GMST table above comes from installing both `astropy-iers-data` versions in
      scratch environments and comparing directly.

## 🏗️ Checklist

## The cost, stated rather than left implicit

**The suite no longer exercises the epoch the examples document.** It cannot catch an
error that depends on being far outside the IERS table — which is the situation this
deliberately stops testing. That trade-off is written into the `_ALIGNED_START`
docstring so the next reader meets it there rather than deducing it.

## A note on the recorded environment

The references record gwmock-signal 0.13.0, which is what this branch resolves. An
earlier pass generated them under 0.14.0 while I was testing the unreleased
`pre_coalescence_duration` API on the same branch. Every comparison still passed —
confirming that addition changed no generated data — but the recorded environment
would have named a version never installed here, so they were regenerated. The
inspiral-cropping fix that needs 0.14.0 is a separate PR.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved stability of simulated signal overlays by aligning event and reference times with a more reliable time period.
  - Updated test data locations for more consistent end-to-end validation.

- **Tests**
  - Refreshed expected noise, signal, waveform, and batched execution results.
  - Updated verification data to reflect the latest simulation environment and regenerated outputs.
  - Preserved validation of noise and signal statistics across supported scenarios.

### DIFF
--- a/tests/e2e/data/bbh_population.csv
+++ b/tests/e2e/data/bbh_population.csv
@@ -1,0 +1,2 @@
+detector_frame_mass_1,detector_frame_mass_2,coa_time,distance,declination,right_ascension,polarization_angle,inclination
+30.0,25.0,1419724820.0,400.0,0.3,1.1,0.2,0.5
--- a/tests/e2e/overlay.py
+++ b/tests/e2e/overlay.py
@@ -24,6 +24,7 @@ from pathlib import Path
 from typing import Any
 
 _REPO_ROOT = Path(__file__).resolve().parents[2]
+_THIS_DIRECTORY = Path(__file__).resolve().parent
 EXAMPLES_DIR = _REPO_ROOT / "examples"
 
 #: Seed pinned for every run, so two runs of one entry are comparable.
@@ -36,7 +37,11 @@ _TEST_SEED = 20260731
 
 #: In-repo CBC population, used in place of every remote population file. Small and fixed, so a
 #: run is reproducible and needs no network.
-POPULATION_FIXTURE = EXAMPLES_DIR / "signal" / "bbh_population.csv"
+#:
+#: Test-owned rather than the example's copy, because its coalescence time has to sit at the epoch
+#: below and the examples deliberately use 2030. Identical to
+#: ``examples/signal/bbh_population.csv`` apart from that time.
+POPULATION_FIXTURE = _THIS_DIRECTORY / "data" / "bbh_population.csv"
 
 #: In-repo pulsar catalogue for the continuous-wave entry.
 #:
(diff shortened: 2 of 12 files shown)


isaac-cf-wong:
@coderabbitai full review

PR #310, opened 2026-08-04
chore(deps): drop the ripple pin, take the floor from gwmock-signal

isaac-cf-wong:
ripplegw 0.3.1 carries the geocentre fix (GW-JAX-Team/ripple#141), and gwmock-signal
0.15.0 floors `ripplegw>=0.3.1` itself. Both change how this repository should express the
constraint, so the development pin can go.

## The problem the pin solved

The continuous-wave simulator generates polarizations at the geocentre so one projection
route serves every source type. Released ripplegw 0.3.0 evaluates the detector latitude as
`arccos(lz / rd)`, which is 0/0 there, and returns NaN for every sample. gwmock-signal
refuses that rather than writing it, so continuous waves simply did not work in this
repository's development or CI without a pinned ripple.

## What changed

Three things together, because any one alone resolves ripplegw 0.3.0 again:

- the `[tool.uv.sources]` table and its rationale are deleted;
- the `jax` and `cuda` extras move to `gwmock-signal[jax]>=0.15.0`;
- the explicit `ripplegw>=0.3` declaration in those extras is **removed** rather than raised
  to 0.3.1.

That last one is the substantive choice. The declaration existed only so the source table
would apply -- uv honours a source only for a package the project declares -- and with the
table gone its only remaining effect would be to restate a version gwmock-signal already
states. The floor belongs beside the guard that raises on the NaNs; two repositories naming
the same version is how the two drift apart.

## Verification

The floor still binds, rather than 0.3.1 merely being latest. A scratch project on
`gwmock-signal[jax]>=0.15.0` plus `ripplegw==0.3.0` is unsatisfiable:

> Because gwmock-signal[jax]==0.15.0 depends on ripplegw>=0.3.1 [...] your project's
> requirements are unsatisfiable.

So the constraint is carried in released metadata, not inferred from resolution order. The
relocked file takes ripplegw 0.3.1 and gwmock-signal 0.15.0 from the registry and no longer
contains any git URL.

Runtime rather than resolution alone: the continuous-wave e2e example runs against registry
ripple, so the NaN is gone in fact and not only by version number. All six
`signal/cw/et_triangle_sardinia` e2e tests pass, including the stored-reference and
reproducibility comparisons.

e2e: 46 passed, 9 skipped. Unit and integration: 971 passed, 23 skipped.

## Summary by CodeRabbit

* **Documentation**
  * Added guidance explaining the minimum supported version and geocentre-related fix.

* **Chores**
  * Updated optional JAX and CUDA setup requirements.
  * Removed temporary development configuration and outdated package declarations.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -45,19 +45,17 @@ sgwb = ["gwmock-signal[sgwb]>=0.13.0"]
 # installed this way batches the work and runs it on the CPU. That is a supported configuration --
 # it is the one the test suite exercises -- but it is not device execution, and nothing at runtime
 # distinguishes the two beyond `jax.devices()`. Use the `cuda` extra below for a GPU.
-# `ripplegw` is named explicitly so the development pin at the bottom of this file applies
-# to it. uv honours a source only for a package the project *declares*, and ripple would
-# otherwise arrive transitively through `gwmock-signal[jax]`; a source alone left the lock
-# resolving ripplegw 0.3.0 from the registry. Declaring it here rather than in a dev group
-# keeps it on the extras that need it -- a dev group would put ripple and JAX into every
-# plain `uv sync`, which would un-skip the ripple-gated tests in the default CI job and make
-# that job quietly test something other than what it claims. The published wheel gains only
-# `ripplegw>=0.3` under these extras, a registry floor that was already transitive, and no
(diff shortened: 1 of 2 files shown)


isaac-cf-wong:
@coderabbitai review

PR #311, opened 2026-08-04
fix(orchestration): claim an event by where its waveform starts

isaac-cf-wong:
A compact binary's buffer begins seconds before its coalescence. Segments claimed an event
by `coa_time`, so an event landing just past a boundary was generated during the *next*
segment -- whose start is after the waveform's. Injection cropped the difference and it
could never be placed, because the earlier segments were already written.

Segments now claim the events whose waveform starts before the segment ends, taken from
`CBCSimulator.pre_coalescence_duration` (gwmock-signal 0.15.0). The existing
forward-overflow carry delivers the remainder, so only the claiming rule changes.

## How much was being lost

Measured for a 30+25 solar-mass binary at 1024 Hz with 16-second segments, against where
`coa_time` lands past the boundary. The lead is 3.5996 s, so 22.5% of uniformly distributed
events lose something:

| coa_time - boundary | dropped | samples | energy |
|---|---|---|---|
| +0.001 s | 3.599 s | 90.0% | 99.78% |
| +0.1 s | 3.500 s | 87.5% | 50.30% |
| +0.25 s | 3.350 s | 83.7% | 11.47% |
| +1.0 s | 2.600 s | 65.0% | 0.34% |
| +3.5 s | 0.100 s | 2.5% | 0.02% |

The energy column collapses because a compact binary's energy sits in the last fraction of a
second before merger, so the damage depends steeply on the offset -- an event 1 ms past a
boundary loses essentially all of its signal.

## Verification

The predicted lead is checked against the buffer generation actually produces, read from the
resulting `TimeSeries.start_time` rather than from another prediction: 3.599609375 s both,
difference exactly 0.0. If the query understated the lead, a segment chosen from it would
still crop.

The payoff is measured as strain energy, not as the absence of a log line. Summed squares
across two assembled segments equal the standalone waveform to rel 1e-9. On the unfixed rule
the same run reports `Discarding 2.600 s (2662 samples)` -- the predicted 2.6 s.

Mutation-tested: reverting to the `coa_time` rule fails 4 tests, removing the warn-once
dedupe fails 1, letting the query failure propagate fails 2.

## Two decisions worth stating

**A failing query does not fail the run.** The boundary test now needs a full parameter set
where before it needed only `coa_time`, so an incomplete catalogue column would crash in
placement. It falls back to the `coa_time` rule and warns once instead. The failure is not
hidden: the same parameters reach generation immediately afterwards, which raises whatever
the query raised, with the context of the event rather than of a boundary test.

**Attribution follows placement.** An event is now listed in the metadata against the frame
its waveform *starts* in, not the frame holding its coalescence, so `find_signal` reports
that earlier frame. Keeping the old key would mean deferring the injection record across a
frame, which leaves no provenance at all for data already written if a run stops between the
two. Both convention comments are updated and a test pins it.

`_report_content_before_segment` stays: a waveform beginning before the run itself has no
segment to go in, and the fallback above still crops. Its docstring and message no longer
claim this is what ordinarily happens.

## Stacking

Targets #310 out of necessity rather than tidiness. With that pin present ripplegw resolves
to `0.3.1.dev3`, below the `>=0.3.1` gwmock-signal 0.15.0 requires, so 0.15.0 -- and this
query -- cannot be installed at all. Merge #310 first; this then retargets to main.

983 passed, 23 skipped. e2e unchanged at 46 passed, 9 skipped: its fixture event sits 4 s
into a single-segment run, so no placement moves and the stored references still match.

## Summary by CodeRabbit

* **Bug Fixes**
  * Waveform segments are now placed and consumed according to waveform start time, improving event ordering and resume behavior.
  * Added fallback handling and warnings when waveform lead-time information is unavailable.
  * Batched and per-event processing now apply consistent placement rules.

* **Metadata**
  * Updated the metadata schema to version 1.4.0 to reflect revised signal-injection attribution.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -22,7 +22,12 @@ classifiers = [
 requires-python = ">=3.12"
 dynamic = ["version"]
 dependencies = [
-  "gwmock-signal>=0.13.0", # 0.13.0 is the first release carrying the continuous-wave simulator
+  # 0.15.0 is the first release exposing `CBCSimulator.pre_coalescence_duration`, which segment
+  # placement asks before generating so an event is claimed by the segment its *waveform* starts in
+  # rather than the one holding its coalescence. Below this the query is absent, placement falls back
+  # to `coa_time`, and the start of any waveform landing near a boundary is cropped -- so this is a
+  # floor the behaviour depends on, not a convenience.
+  "gwmock-signal>=0.15.0",
   "gwmock-noise>=0.10.0",
   "gwmock-pop>=0.11.4",
   "typer>=0.19.0",         # Literal-type CLI options; the SPEC 0 floor (0.13) predates them
--- a/src/gwmock/cli/adapter_orchestration.py
+++ b/src/gwmock/cli/adapter_orchestration.py
@@ -149,9 +149,16 @@ class AdapterOrchestrator(TimeSeriesMixin, Simulator):
         # Recorded into per-batch metadata so a frame's sources are self-describing.
         self._batch_injections: list[dict[str, Any]] = []
         self._pending_noise_chunk: dict[str, Any] | None = None
-        # Warn once per run, not once per event: a catalogue the query cannot answer for cannot
-        # answer for any of its events, and the message would otherwise arrive thousands of times.
-        self._pre_coalescence_query_failed = False
+        # Warned at most once per distinct reason, not once per event: a catalogue whose parameters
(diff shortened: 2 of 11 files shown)


isaac-cf-wong:
@coderabbitai review

isaac-cf-wong:
@coderabbitai review

PR #312, opened 2026-08-04
docs(tests): state the cutoff behind the placement loss figures

isaac-cf-wong:
## 📝 Summary

The energy-loss curve in the placement test's docstring — 0.34% at 1 s past a boundary rising
to 99.8% at 1 ms — was measured at this module's 30 Hz cutoff and did not say so. Same omission
as the one fixed in Leuven-Gravity-Institute/gwmock-signal#369, made one commit later in this
repository's own work.

It matters because 30 Hz is the mild end. The same binary and network at a 20 Hz cutoff, which
is what a real ET configuration uses, loses 32.9% at 0.5 s past the boundary against 0.92% at
30 Hz. LAL's conditioning rounds both cutoffs to the same 4.000 s buffer, so the dropped *span*
is identical at 3.100 s and only its content differs. Quoted without its cutoff the curve
understates the defect the test exists for by a factor of 35.

Two corrections from review:

- The cross-reference quoted 32.3% and 0.91%, which are **single-detector** values, inside a
  module that configures the ET-Triangle-Sardinia network. Now 32.9% and 0.92%, measured on the
  network the module actually uses.
- The rounding coincidence is a property of the **chirp-time bin**, not of "these masses": it
  holds for a range of heavier pairs and breaks for lighter ones, so the 10+10 counterexample
  (16 s at 20 Hz against 8 s at 30 Hz) is now given inline.

Also stopped rounding 11.47% to 11% and 50.30% to 50%, which cost precision for nothing and
made the figures harder to reconcile with the same numbers elsewhere.

The full tables live in gwmock-signal's user guide at
`user_guide/segment-placement-losses.md`; this docstring points there rather than duplicating
them.

## ✨ Type of Change

- [x] 📝 Documentation update

- [x] **Unit Tests:** `uv run --extra jax pytest tests/cli/test_inspiral_segment_placement.py` — 24 passed
- [x] **Every figure re-measured** rather than copied: 0.338% at 1 s, 11.469% at 0.25 s, 50.302%
      at 0.1 s, 99.778% at 1 ms on ET at 30 Hz; 32.93% and 0.9191% at 0.5 s for the 20/30 Hz
      comparison.

## 🏗️ Checklist

- [ ] I have added tests that prove my fix is effective or that my feature works — not
      applicable: a test docstring only, no behaviour change.

Not a UI change.

## Summary by CodeRabbit

* **Documentation**
  * Expanded test documentation with details on 30 Hz and 20 Hz configurations.
  * Added explanations of energy loss across offsets, buffer rounding, and backend or mass variability.
  * Documented the ET-Triangle-Sardinia network used in the tests.

### DIFF
--- a/tests/cli/test_inspiral_segment_placement.py
+++ b/tests/cli/test_inspiral_segment_placement.py
@@ -451,7 +451,7 @@ class TestNothingIsDiscarded:
 
         This offset is deliberately the *mild* case. **At this module's 30 Hz cutoff, on the
         ET-Triangle-Sardinia network it configures**, the unweighted strain-squared energy lost runs
-        0.34% at 1 s past the boundary, 11% at 0.25 s, 50% at 0.1 s and 99.8% at 1 ms, because a
+        0.34% at 1 s past the boundary, 11.5% at 0.25 s, 50.3% at 0.1 s and 99.8% at 1 ms, because a
         compact binary's energy is concentrated in the last fraction of a second before merger. Testing the mild case makes the assertion depend on placement rather
         than on that concentration: 0.34% is a loss no energy-fraction tolerance would forgive, and a
         test pinned at the catastrophic end would still pass if placement were only approximately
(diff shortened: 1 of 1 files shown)


isaac-cf-wong:
@coderabbitai full review

PR #313, opened 2026-08-04
test(assembly): pin that the two segment assemblers agree

isaac-cf-wong:
## 📝 Summary

Two functions in this ecosystem place signals into segments, and a wrong segment looks exactly like
a right one. `gwmock_signal.jax_batch.assemble_segments` scatters an in-memory batch into a known
tiling; gwmock streams, holding one segment at a time, claiming events for it and carrying overflow
forward as a tail. This closes the concern that the two would drift silently.

**Ownership was already decided, just unguarded.** gwmock does not call `assemble_segments`
anywhere — the batched path is a different way to *generate*, not to assemble, which
`_simulate_batched_segment` already documents. They are not duplicate implementations of one rule but
two regimes, so the fix is an equivalence guard rather than a deduplication.

**They agree, with a stated bound rather than exactly.** The residue is a *generation* artifact, not
an assembly difference: a JAX batch is padded to its longest event, so gwmock (per-segment batches)
and the scatter (one catalogue batch) compute the same event at different batch shapes and the FFTs
differ in their last bits — 4,608 samples for a 30+25 binary alone, 236,196 once a 1.4+1.4 binary
joins it. Measured at 3.3e-13 of peak for a uniform catalogue and 2.0e-11 with a binary neutron star
mixed in, against a 1e-9 tolerance.

An earlier revision of this PR blamed the sampling-grid anchor. That was refuted by measurement:
generating the same event on a run-start grid and a segment-start grid is bit-identical.

**An unplanned check on #311.** Reverting to the old `coa_time` claiming rule makes gwmock disagree
with the in-memory assembler by **17.7% of peak** in the first segment, and report holding 98.8% of
the energy the generator produced. The two agree only because of that fix; the in-memory assembler
never needed a claiming rule, so it was right all along and gwmock has caught up.

## What is tested

- **per-sample equivalence**, relative to each segment's peak with `atol=0.0` — strain is ~1e-21 and
  its square ~1e-42, so the default absolute tolerance would make any two such arrays compare equal;
- **conservation against an independent oracle** — the energy generation produced inside the run
  window, touching neither assembler, so a failure says *which* one is short and by how much. Scope
  stated in the code: it cannot see loss the two share, because an event whose buffer begins before
  the run is truncated identically by both and by the oracle. That loss is real but is not an
  assembly defect, and `_report_content_before_segment` reports it per signal;
- **a mixed-mass catalogue**, because the uniform one is the easy case and batch padding is where
  agreement could plausibly break;
- **ownership**, by instrumenting the real `assemble_segments` and observing calls during a gwmock
  run — with a positive control first, since a guard watching the wrong name would report "not
  called" forever;
- **a boundary where the two are *not* equivalent**: 16.0006 s at 1024 Hz is 16384.6144 samples, so
  gwmock's `int()` truncates to exactly 16.0 s of data while the scatter raises. Pinned so a change
  making the scatter round silently does not turn a loud configuration error into two subtly
  different datasets.

## ✨ Type of Change

- [x] 🎨 Refactoring (no functional changes, no API changes) — test-only addition

- [x] **Unit Tests:** `uv run --extra jax pytest tests/cli/test_assembly_ownership.py` — 5 passed
- [x] **Full suite:** 998 passed, 23 skipped
- [x] **Mutation-tested:** reverting the claiming rule fails both the per-sample and conservation
      assertions, the first naming 17.7% of peak and the second the −1.1967% energy gap.
- [x] **Two review rounds**, which refuted the first revision's stated mechanism, showed its
      equivalence claim was tested only where every buffer is the same length, and proved its
      conservation assertion mathematically redundant (per-sample equality implied it). All three are
      addressed above.

One review figure was **not** adopted: a reported mixed-mass divergence of 5.85e-4 of peak through
the pipelines. Measured here as 2.0e-11 — the 5.86e-4 is a per-event *buffer* comparison, which does
not survive assembly at that magnitude. The mixed case is now a test, so CI settles it.

## 🏗️ Checklist

- [ ] I have made corresponding changes to the documentation — not applicable; the ownership decision
      was already documented in `_simulate_batched_segment`, and this references it.

Not a UI change.

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive validation to ensure streamed and scattered assembly produce matching strain results.
  * Added checks for generated energy conservation, mixed-mass catalogues, and off-grid duration boundaries.
  * Added safeguards confirming assembly behavior remains independent and consistent.
  * Tests requiring Ripple are automatically skipped when it is unavailable.

### DIFF
--- a/tests/cli/test_assembly_ownership.py
+++ b/tests/cli/test_assembly_ownership.py
@@ -1,0 +1,217 @@
+"""Two implementations could place signals into segments; this pins that they agree.
+
+``gwmock_signal.jax_batch.assemble_segments`` scatters a batch into a fixed tiling of segments in
+memory, given every segment start up front. gwmock assembles differently by necessity: it streams,
+holding one segment at a time, claiming events for it and carrying the overflow forward as a tail.
+
+gwmock owns assembly and does not call ``assemble_segments`` -- the batched path is a different way
+to *generate*, not a different way to assemble. That decision is recorded in
+:meth:`~gwmock.cli.adapter_orchestration.AdapterOrchestrator._simulate_batched_segment`, and it is
+the reason this file exists rather than a deduplication: two assemblers stay in the ecosystem, so
+the risk is that they drift while both keep producing plausible segments. Nothing about a wrong
+segment looks wrong.
+
+So the guard is an equivalence measurement rather than a shared implementation. Same events, same
+generation, both assemblers, compared sample by sample.
+"""
+
+from __future__ import annotations
+
+import csv
+from pathlib import Path
+from typing import Any
+
+import numpy as np
+import pytest
+
+from gwmock.cli.utils.config import Config
+
+_FS = 1024.0
+_SEGMENT = 16.0
+_SEGMENTS = 3
+_START = 1577491296.0
+_DETECTOR = "ET-Triangle-Sardinia"
+
+#: One event mid-segment, one just past a boundary so the claiming rule has to pull it earlier, and
+#: one in the last segment. The middle event is the interesting one: gwmock claims it for the
+#: *previous* segment because its waveform starts there, while ``assemble_segments`` simply scatters
+#: it into both. If the two disagree anywhere, it is on an event like that.
+_EVENTS: list[dict[str, Any]] = [
+    {
+        "detector_frame_mass_1": 30.0,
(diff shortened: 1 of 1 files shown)


isaac-cf-wong:
@coderabbitai full review

PR #314, opened 2026-08-05
perf(population): transfer a population in one copy instead of per sample

isaac-cf-wong:
## 📝 Summary

`gwmock-pop` returns **JAX device arrays**. `PopulationAdapter.__init__` stored them with
`tuple(values)`, and iterating a device array pulls each element back individually — one device
dispatch per sample. Now `tuple(np.asarray(values).tolist())`: one bulk transfer, then plain Python
floats.

It made *iteration* slow too, less obviously: the resulting tuple held JAX scalars, so
`_coerce_event_value`'s `.item()` was another device call per parameter per event.

**Steady state, both paths warmed, 500 events × 8 parameters: 24.2 ms against 0.135 ms — a factor of
180.** An independent re-measurement got ~220×, and 37× per event.

An earlier revision of this PR quoted 42,000× and 469×. Those mixed a *cold* `tuple(jax)` against a
*warm* `tolist`. Per-element `.item()` swings 13–397 µs with JAX warm state, so the "before" side is
unstable by one to two orders of magnitude and does not belong in a headline. 180× is the figure that
reproduces.

`tolist` rather than keeping the array, because indexing an array builds a numpy scalar each time —
15.4 µs per event against 1.4 µs for a tuple of floats. The bulk form is used for the transfer and
discarded.

The container stays a tuple, but the **elements change** from JAX or numpy scalars to Python ones.
Event dictionaries are unaffected, since `_coerce_event_value` already called `.item()`, but
`population_mapping` exposes different element types. That is intended, and the tests assert it.

## A pre-existing bug, fixed here

`tuple(values)` on a 2-D array yields row arrays, and the one-dimensional check in
`_validate_parameter_values` looks for a `.shape` that a tuple does not have. So a (3, 2) column
silently became a 3-event catalogue of array-valued parameters and nothing refused it. The array is
now passed through so the validator sees its shape. Both reviewers found this independently;
mutation-testing the old behaviour gives `DID NOT RAISE ValueError`.

## Coverage

Codecov flagged 66.66% patch coverage. Now **100% statement and branch** for this module (89
statements, 22 branches).

Reaching that removed code rather than adding it. The `dtype == object` fallback I had written **did
nothing**: its test passed with the branch deleted, so the test could not fail, and `tolist` on an
object array returns the objects unchanged — checked for `None`, dicts, `Decimal`, `datetime` and
nested lists. Deleted, so the diff is now smaller than before.

The validator's `TypeError` branch turns out to be unreachable through the public API: everything
`_materialise_on_host` returns is a tuple or an array, so anything not one-dimensional meets the shape
check first. It is tested at the private boundary with that stated, rather than presented as a
reachable path.

## ✨ Type of Change

- [x] 🐛 Bug fix (non-breaking change which fixes an issue) — the 2-D column silently reinterpreted
- [x] 🎨 Refactoring (no functional changes, no API changes) — the performance change itself

- [x] **Unit Tests:** `uv run --extra jax pytest tests/population/` — 25 passed, and 6 consecutive
      clean runs after a reviewer saw one transient failure while the tree was mid-edit
- [x] **Full suite:** 1008 passed, 23 skipped
- [x] **Values, across both reviewers:** old and new produce bit-identical results for float16/32/64,
      bfloat16, complex, int, uint, bool, weak types, NaN, inf, signed zero, subnormals and extremes.
      Production is float64 (`gwmock_pop` enables x64 on import), so `coa_time` round-trips exactly at
      GPS scale; under `JAX_ENABLE_X64=0` it would be float32 and lossy, which is pre-existing.
- [x] **The guard counts conversions rather than inferring them.** The first version asserted the
      stored elements are Python floats; a reviewer defeated it with a per-element implementation that
      also leaves Python floats, passes every test, and costs 1.4 s per 500 elements. A spy now counts
      bulk conversions and element reads, and fails that counterexample with
      `read 5 elements individually`.
- [x] **Which tests guard what is stated in the file.** The value-preservation and object-column tests
      check behaviour and pass under a revert; the conversion spy, the ragged fallback and the 2-D
      refusal are the discriminating ones.

## 🏗️ Checklist

- [ ] I have made corresponding changes to the documentation — not applicable; internal storage change
      with no API or configuration surface.

Not a UI change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved population data handling for bulk conversion between backend and host environments.
  * Preserved numeric values and scalar types, including ragged or object-based inputs.
  * Added validation for invalid data dimensions and unsupported inputs.
  * Maintained existing validation behavior while avoiding unnecessary element-level reads.

### DIFF
--- a/src/gwmock/population/adapter.py
+++ b/src/gwmock/population/adapter.py
@@ -27,9 +27,10 @@ def _materialise_on_host(values: Sequence[Any]) -> tuple[Any, ...]:
     a tuple of floats, measured over eight parameters. So the bulk form is used for the transfer and
     discarded, which also keeps the stored type exactly what it was before.
 
-    Anything numpy cannot represent as a numeric array -- object columns, ragged input -- falls back
-    to the plain tuple. The per-element cost is real there, but such values were never
-    device-resident to begin with.
+    Ragged input, which ``np.asarray`` refuses, falls back to the plain tuple. Object columns do
+    *not* need a fallback: ``tolist`` on an object array hands the objects back unchanged, checked for
+    ``None``, dicts, ``Decimal``, ``datetime`` and nested lists, so a branch for them would be code
+    that cannot change an outcome.
 
     Args:
         values: One parameter's values, as the backend produced them.
@@ -40,8 +41,6 @@ def _materialise_on_host(values: Sequence[Any]) -> tuple[Any, ...]:
     try:
         materialised = np.asarray(values)
     except (TypeError, ValueError):
-        return tuple(values)
-    if materialised.dtype == object:
         return tuple(values)
     if materialised.ndim != 1:
         # Returned as the array, deliberately, so the shape check in `_validate_parameter_values`
--- a/src/gwmock/population/adapter.py
+++ b/src/gwmock/population/adapter.py
@@ -6,7 +6,40 @@ from collections.abc import Iterator, Mapping, Sequence
 from types import MappingProxyType
 from typing import Any
 
+import numpy as np
 from gwmock_pop import GWPopSimulator
+
+
+def _materialise_on_host(values: Sequence[Any]) -> tuple[Any, ...]:
+    """Return *values* as a tuple of plain Python scalars, transferred in one go.
+
+    ``gwmock-pop`` hands back **JAX device arrays**, and iterating one -- which a bare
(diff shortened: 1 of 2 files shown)


isaac-cf-wong:
@coderabbitai full review

PR #318, opened 2026-08-06
fix(provenance): record a signal against every frame it reaches

isaac-cf-wong:
## 📝 Summary

`gwmock find-signal` named one frame when a signal's samples are in several.

Measured with a 1.6+1.4 binary from 30 Hz across 32 s segments, whose waveform spans three frames (nine files across three detectors): the lookup returned **three of nine**, and the six it omitted included the **merger** — the named frame held a peak of 1.29e-23 against 7.28e-23 in one it did not name. For continuous waves it is universal, since every pulsar is in every frame of the run.

Both lookup paths were affected. Parameter lookup is documented as the source of truth, but for a compact binary the later batches recorded `injections: []`, so it returned the same three frames.

Two mechanisms, which fail independently:

1. Chunks carry `event_id` beside `injection_parameters`, and `TimeSeriesMixin.simulate` collects the contributing records **before** injecting. Injection sums chunks into shared channels, so afterwards there is no way to ask which signal put samples where — it has to be asked first. `_segment_injections` unions what a batch generated with what reached it, deduplicated on `event_id` because a multi-detector batch emits one chunk per detector.
2. `signal_index.yaml` accumulates instead of assigning. Contributions are stored per batch rather than as a flat frame list, because a re-run has to withdraw its own contribution and a flattened copy alongside would be a second place for the same fact to live; flattening happens once, on read. Pre-1.5.0 entries are migrated on read rather than rejected — an index is a rebuildable cache, and refusing one would make an upgrade look like data loss.

| | before | after |
|---|---|---|
| CBC `--id` | 3 of 9 frames, missing the merger | 9 of 9 |
| CBC `--param` | 3 of 9 | 9 of 9 |
| CW `--id` | 3 of 9 | 9 of 9 |

`SCHEMA_VERSION` 1.4.0 → 1.5.0. `injections` kept its type while changing its cardinality, so a consumer summing it across a run will overcount and only the version says so. `find_signals` reports `metadata` as a list on both paths.

## ✨ Type of Change

- [x] 💥 Breaking change — `signal_index.yaml` changed shape and `injections` changed meaning. Old indexes are read; external raw-YAML parsers of the index will need updating.
- [x] 📝 Documentation update

**Run the suite on `macmini`, not the primary machine (4 cores).** Use your own checkout per `_instructions/test-host-policy.md`. Two traps: `uv` is at `~/.local/bin/uv` and not on the non-interactive SSH PATH; and **do not use `uv sync --all-extras`** — the `cuda` extra has no macOS wheel, so the sync fails, pytest never runs, and a failure count of 0 on the missing log reads exactly like a pass. Use `uv sync --extra jax --extra sgwb`.

- [x] Full suite: 1029 passed on Linux, 1029 on macmini.
- [x] End to end through the CLI for both source types, table above.
- [x] **Both mechanisms mutation-checked.** Reverting the index to assignment fails `test_one_event_reaching_three_batches_names_all_their_frames`; zeroing the carry-forward collection fails `test_a_signal_crossing_a_boundary_is_recorded_against_both_segments` with `assert [] == [0]`; dropping the write-side migration fails `test_a_new_batch_does_not_discard_a_pre_migration_entry`.
- [x] `contributes_samples` is parametrised across the boundary, including chunks ending exactly at the segment start and starting on its last sample, where an off-by-one costs exactly one frame per signal.

Worth reviewers' attention:

- **The wiring, not the helper, was where the defect lived.** A first pass tested `_contributing_injections` thoroughly and left `simulate` uncovered — the helper can be entirely correct with nothing calling it. Only an end-to-end mutation exposed that, which is why one test goes through the orchestrator.
- **An existing test asserted the opposite of this change and kept passing**, because it reads `_batch_injections` (what a batch generated) rather than what reaches the frame. True of that list, false of the metadata; rewritten to pin the distinction rather than deleted.

## 🏗️ Checklist

## Known limitations, named rather than fixed

Both predate this work, both have tracker items, and neither is a provenance bug:

- **Checkpoint resume loses spillover** (`gwmock/spillover-lost-on-checkpoint-resume`). `cached_data_chunks` is not a `StateAttribute`, so a resumed run regenerates neither the tail nor its record: through the CLI, segment 1's peak goes from 8.6e-22 to **0.0** after a resume — the merger is gone. The samples vanish whether or not anything records them, so fixing it here would mix a data fix into a metadata one. It does make this PR's claim false at exactly a resume boundary, which is documented where the claim is made.
- **The index has no write lock** (`gwmock/signal-index-concurrent-writers`). Unlocked read-modify-write, reproducible with a barrier. Not made worse here, but a dropped update used to cost one assignment and now costs one batch's frames for every event in it. Bounded in practice: batches are written sequentially within a run, so it bites only when two runs share a metadata directory.

**Not verified:** `execution: batched` across a chunk boundary at production scale, and what any existing external consumer of `signal_index.yaml` does with the new shape.

## Summary by CodeRabbit

* **New Features**
  * Signal provenance now appears in every frame containing the signal, including signals crossing segment boundaries.
  * Signal indexes preserve contributions across batches and provide associated frame and metadata details.
  * Signal lookups consistently return metadata files as a list.
  * Provenance includes event identifiers and deduplicates repeated records.

* **Compatibility**
  * Existing pre-1.5.0 signal indexes remain readable.

* **Documentation**
  * Updated reproducibility and signal lookup guidance to explain multi-frame signal behavior and schema 1.5.0.

### DIFF
--- a/docs/user-guide/reproducibility.md
+++ b/docs/user-guide/reproducibility.md
@@ -5,12 +5,12 @@
 
 ## Schema
 
-Each record is validated at write time and uses schema version `1.3.0`.
+Each record is validated at write time and uses schema version `1.5.0`.
 Consumers must reject unknown major versions.
 
 ```json
 {
-    "schema_version": "1.3.0",
+    "schema_version": "1.5.0",
     "gwmock_version": "x.y.z",
     "subpackage_versions": {
         "gwmock_signal": "x.y.z",
--- a/docs/user-guide/reproducibility.md
+++ b/docs/user-guide/reproducibility.md
@@ -95,9 +95,13 @@ gwmock rewriting their internal structure.
 `signal.injections` records the source parameters of the signals attributed to
 that batch's frame(s), in injection order. Each entry is
 `{"event_id": <index in the population>, "parameters": {...}}`. A signal is
-attributed to the frame in which it **merges** (its `coa_time` falls inside that
-frame's time segment); a long waveform whose inspiral or ringdown extends into
-adjacent frames is listed only under its merger frame. `event_id` is the event's
+listed against **every frame its samples reach**, so a long inspiral crossing a
+segment boundary appears in each frame it spans, and a continuous wave appears
+in all of them. This changed in schema 1.5.0: a signal used to be listed only
+under the frame it was generated for, which for a 48 s inspiral across 32 s
+segments meant one frame out of three -- and not the one holding the merger. The
+frame a signal is _generated_ for is the one its waveform **starts** in (schema
+1.4.0; before that, the one its `coa_time` fell in). `event_id` is the event's
 index in the population as ordered for the run (by `coa_time` under the default
 ordering), so it is stable for a fixed configuration. Stationary/SGWB segments
 have no discrete events and record an empty list.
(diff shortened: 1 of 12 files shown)


isaac-cf-wong:
@coderabbitai review

PR #319, opened 2026-08-06
fix(checkpoint): carry spillover across a resume so the tail is not lost

isaac-cf-wong:
## 📝 Summary

A run resumed from a checkpoint lost the tail of any signal crossing the resume point — the **samples**, not just the provenance. `cached_data_chunks` holds the part of a chunk extending past the segment being built, and it existed only in memory, so a resumed process started with none of it.

Same procedure both sides — 1.6+1.4 M☉ from 30 Hz across 32 s segments, interrupt once `.gwmock_checkpoints/simulation.checkpoint.json` appears, re-run:

| frame | uninterrupted | resumed on `main` | resumed with this |
|---|---|---|---|
| ...540 | 1.286e-23 | 1.286e-23 | 1.286e-23 |
| ...572 | 1.837e-23 | 1.837e-23 | 1.837e-23 |
| ...604 (merger) | 7.280e-23 | **0.000e+00** | 7.280e-23 |

The frames before the resume point are bit-identical, so nothing about the run looks wrong — the loudest part of the signal is simply absent.

**Two defects, and fixing either alone leaves a hole.** Besides the chunks being unreachable, `TimeSeries.to_json_dict` carried neither `metadata` nor channel identity, so restoring them would have restored samples that no longer say which signal they are — and `inject` copies the channel name onto a tail deliberately, for the same reason.

**Why spillover is not a `StateAttribute`**, which is the obvious fix and does not work: `state` is serialized into every *batch metadata record* as well as the checkpoint, so it would write megabytes of samples into a provenance document meant to stay small, and those records are dumped with plain `json`, which has no encoder for a `TimeSeriesList`. The whole suite fails with exactly that error, which is how the constraint was found. Spillover therefore travels as its own checkpoint field, `last_simulator_spillover`, **scoped to the simulator and batch that produced it** — a plan can execute several simulators, and handing one's tail to another is worse than losing it: real strain of the right shape at the wrong time, which nothing downstream would flag.

**Serialization cost, measured.** `.tolist()` bypassed the encoder's base64 path, which did not matter until spillover started going to disk after every batch. At 100 s × 3 detectors, 4096 Hz: 44.1 MB and 9.02 s as lists against 13.1 MB and 0.43 s as base64. Measured directly at a realistic 1000 s BNS tail rather than extrapolated: **131 MB and 1.1 s, 1.33× the raw float64 bytes**. Round-trip is exact, checked with `atol=0.0`. The checkpoint is also read once per plan rather than per batch, since re-reading 131 MB in the loop would pay that every time.

## ✨ Type of Change

Checkpoints gain a field; one written before this reads back as no spillover rather than raising, so an upgrade mid-run does not cost the run.

**Run on `macmini`, not the primary machine (4 cores).** Own checkout per `_instructions/test-host-policy.md`. Two traps: `uv` is at `~/.local/bin/uv`, not on the non-interactive SSH PATH; and **do not `uv sync --all-extras`** — the `cuda` extra has no macOS wheel, the sync aborts, pytest never runs, and a failure count of 0 on the missing log reads exactly like a pass. Use `uv sync --extra jax --extra sgwb`, and check for the summary line.

- [x] Full suite: **1036 passed** on Linux and on macmini.
- [x] **End-to-end resume is automated**, not described: the test runs the CLI, waits for the checkpoint *file* rather than a wall-clock delay, interrupts, resumes, and compares against an uninterrupted run in a separate directory. Marked `integration`.
- [x] Every mechanism mutation-checked: stubbing the restore, dropping `metadata` from serialization, and gutting the save call site in `execute_plan` each fail their own test — the last at `0.000e+00 against 1.837e-23`.
- [x] Scoping tested both ways: another simulator's tail is refused, and so is any batch but the immediate successor.

Worth reviewers' attention:

- **The save side was the gap.** Every unit test handed the spillover to `save_checkpoint` directly, so gutting the one call site in `execute_plan` left them all green. A first attempt to close that spied on `save_checkpoint` and then called it directly — it asserts nothing about the call site and passed with the site gutted, so it was deleted rather than kept.
- **Known and not fixed:** dtype widens across the round trip (float32 → float64; costs memory, not accuracy). The other restore branch, which resumes from a batch metadata record, still cannot carry spillover because those records hold no samples by design — stated at the call site. Two independent reviewers disagreed about how reachable that branch is in a real resume, and I have not resolved it.

## 🏗️ Checklist

Not applicable.

## Summary by CodeRabbit

- **Bug Fixes**
  - Resuming an interrupted simulation now preserves pending spillover data, helping produce results consistent with uninterrupted runs.
  - Spillover is restored only for the matching simulator and immediately following batch, including retry scenarios.
  - Existing checkpoints without spillover remain compatible.

- **Improvements**
  - Time series data now preserves channel names, identifiers, units, and metadata during serialization and restoration.
  - Checkpointed time series data uses a more efficient format while retaining available channel information.

### DIFF
--- a/src/gwmock/cli/simulate_utils.py
+++ b/src/gwmock/cli/simulate_utils.py
@@ -725,7 +725,10 @@ def instantiate_simulator(
 
 
 def restore_batch_state(
-    simulator: Simulator, batch: SimulationBatch, last_simulator_state: dict[str, Any] | None = None
+    simulator: Simulator,
+    batch: SimulationBatch,
+    last_simulator_state: dict[str, Any] | None = None,
+    last_simulator_spillover: Any = None,
 ) -> None:
     """Restore simulator state from batch metadata or checkpoint file if available.
 
@@ -775,6 +778,12 @@ def restore_batch_state(
                 last_simulator_state.get("counter"),
             )
             simulator.state = last_simulator_state
+            # Restored only on this branch -- the one that resumes from the checkpoint's *last*
+            # state. The branch above restores from a batch metadata record, which by design does
+            # not carry samples, so there is no spillover to restore there and a run resumed that
+            # way still loses the tail. Stated rather than silently half-handled.
+            if last_simulator_spillover is not None:
+                simulator.cached_data_chunks = last_simulator_spillover
             logger.debug(
                 "[RESTORE] Batch %d: State restored successfully - new_counter=%s",
                 batch.batch_index,
@@ -1223,9 +1232,11 @@ def execute_plan(  # noqa: PLR0915
     if not resuming:
         logger.debug("No checkpoint found or no batches completed yet")
         last_simulator_state = None
+        last_simulator_spillover = None
     elif completed_batch_indices == loaded_batch_indices:
         logger.info("Loaded checkpoint: %d batches already completed", len(completed_batch_indices))
         last_simulator_state = checkpoint_manager.get_last_simulator_state()
+        last_simulator_spillover = checkpoint_manager.get_last_simulator_spillover()
     else:
         # One or more checkpointed batches are missing their outputs. The checkpoint
         # only holds the tail simulator state, so an interior batch cannot be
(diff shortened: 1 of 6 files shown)


isaac-cf-wong:
@coderabbitai review

isaac-cf-wong:
@coderabbitai review

PR #321, opened 2026-08-06
fix(checkpoint): refuse to resume from a checkpoint another run wrote

isaac-cf-wong:
## 📝 Summary

Two configs run from one working directory, the first interrupted after two of its three batches. The second **loaded the first's checkpoint, skipped those batches as already complete, and wrote one frame where a clean run writes three** — exit code 0, no warning.

Silent data loss from an ordinary action: every shipped example puts `.gwmock_checkpoints` and `metadata/` relative to the working directory, so running two configs from one directory is a normal thing to do.

**The guard that exists for this cannot catch it.** `_batch_outputs_present` looks up `{simulator_name}-{batch_index}.metadata.json` — a name with nothing config-specific in it — then verifies the outputs *that file* records. The first run's metadata records the first run's frames, which exist, so the check passes on one run's evidence while a different run is executing.

Checkpoints now carry a **run fingerprint**: every batch's `config_sha256` plus the resolved output and metadata directories. `require_matching_config` refuses a mismatched resume before anything is read from the checkpoint, naming both hashes and the file so the user can see which two runs collided and what to move.

| scenario | before | after |
| --- | --- | --- |
| different config, same directory | 1 frame of 3, exit 0 | 0 frames, exit non-zero |
| same config, different `--output-dir` | 2 frames of 3, exit 0 | 0 frames, exit non-zero |
| legitimate resume of the same run | 3 frames | 3 frames |
| refused run with `--ignore-checkpoint` | — | 3 frames |

**Why the identity is the run, not the config file.** The first version of this fix hashed only the config file, which left the same bug reachable through `--output-dir` — reproduced at 2 frames of 3. Directories are resolved so `./out` and its absolute form stay one run; every batch's hash counts rather than the first, because a plan assembled from several metadata records can mix configurations.

**Why it refuses rather than silently starting fresh.** Ignoring the foreign checkpoint would fix the data loss and hide its cause — a stale checkpoint left in place and a full re-run nobody asked for. Refusing is also the recoverable direction: deleting someone's checkpoint on their behalf is not.

**`--ignore-checkpoint`** exists because a refusal with no way out is a dead end for an automated campaign that cannot answer a prompt. It discards the checkpoint for that invocation and warns; it deletes nothing. The default stays a refusal, so nothing is skipped by accident.

## ✨ Type of Change

- [x] ✨ New feature — `--ignore-checkpoint`

Checkpoints gain a field. One written before this is accepted **with a warning** rather than refused, so upgrading mid-run does not break a legitimate resume.

**Run on `macmini`, not the primary machine (4 cores).** Own checkout per `_instructions/test-host-policy.md`. Two traps: `uv` is at `~/.local/bin/uv`, not on the non-interactive SSH PATH; and **do not `uv sync --all-extras`** — the `cuda` extra has no macOS wheel, the sync aborts, pytest never runs, and a failure count of 0 on the missing log reads exactly like a pass. Use `uv sync --extra jax --extra sgwb` and check for the summary line.

- [x] Full suite: **1050 passed** on Linux and on macmini.
- [x] Every row of the table above reproduced through the CLI.
- [x] Both halves mutation-checked: removing the guard and not saving the fingerprint each fail the end-to-end test. The `--output-dir` hole and the escape hatch each have their own end-to-end test.
- [x] Fingerprint properties unit-tested: output directory, metadata directory, all batch hashes, path equivalence, and stability across identical inputs.

## 🏗️ Checklist

## Known gaps, stated rather than fixed

All three established in review; none is worse than before this branch.

- **Cosmetic config edits refuse a valid resume.** The hash is over config bytes, so adding a comment or reindenting changes it. Safe direction, and `--ignore-checkpoint` is the way out.
- **External inputs are invisible.** The same config file with a different population CSV behind it fingerprints identically, so a resume can mix catalogues. Hashing referenced inputs is the fix if it bites.
- **Legacy checkpoints stay exposed.** A pre-fingerprint checkpoint is trusted with a warning, and interrupted runs are exactly the population that resumes, so this is not a rare corner. It closes as those checkpoints age out.

Separately, `gwmock/signal-index-concurrent-writers` shares the root cause — two runs sharing a directory — but `signal_index.yaml` has no fingerprint and is not gated here. Reviewers judged it lower severity, since the index is a rebuildable cache that `simulate` never consumes.

## Summary by CodeRabbit

* **New Features**
  * Added an option to start simulations without using existing checkpoints.
  * Checkpoints now verify that saved progress matches the simulation configuration and output locations.
  * Compatible checkpoints continue to resume normally, while mismatched checkpoints are rejected with diagnostics.
  * Older checkpoints without configuration details remain supported with a warning.

### DIFF
--- a/src/gwmock/cli/simulate.py
+++ b/src/gwmock/cli/simulate.py
@@ -58,6 +58,7 @@ def _maybe_isolate(
     isolate: bool,
     output_dir: str | None,
     overwrite: bool,
+    ignore_checkpoint: bool = False,
     author: str | None,
     email: str | None,
     dry_run: bool = False,
@@ -103,6 +104,8 @@ def _maybe_isolate(
         child_args += ["--output-dir", output_dir]
     if overwrite:
         child_args.append("--overwrite")
+    if ignore_checkpoint:
+        child_args.append("--ignore-checkpoint")
     if author:
         child_args += ["--author", author]
     if email:
@@ -120,6 +123,7 @@ def _simulate_impl(  # pylint: disable=too-many-locals, too-many-branches, too-m
     output_dir: str | None = None,
     metadata_dir: str | None = None,
     overwrite: bool = False,
+    ignore_checkpoint: bool = False,
     metadata: bool = True,
     author: str | None = None,
     email: str | None = None,
@@ -198,6 +202,7 @@ def _simulate_impl(  # pylint: disable=too-many-locals, too-many-branches, too-m
                 isolate=isolate,
                 output_dir=output_dir,
                 overwrite=overwrite,
+                ignore_checkpoint=ignore_checkpoint,
                 author=author,
                 email=email,
                 dry_run=dry_run,
@@ -280,6 +285,7 @@ def _simulate_impl(  # pylint: disable=too-many-locals, too-many-branches, too-m
                 output_directory=final_output_dir,
                 metadata_directory=final_metadata_dir or Path("metadata"),
                 overwrite=overwrite,
+                ignore_checkpoint=ignore_checkpoint,
                 max_retries=3,
             )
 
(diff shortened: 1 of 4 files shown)


isaac-cf-wong:
@coderabbitai review

PR #323, opened 2026-08-07
fix(index): make signal-index updates safe against concurrent writers

isaac-cf-wong:
## 📝 Summary

`update_signal_index` had two independent defects, both losing data silently with exit
code 0.

**1. Unlocked read-modify-write.** Two runs sharing a metadata directory raced. The
loser's events vanished from `signal_index.yaml` while their samples stayed in the
frames, so `gwmock find-signal` reported nothing for an event that had been injected.

**2. Non-atomic write.** The write truncated the index in place, and the loader treats an
unparsable index as "create a new index" — so a crash or a full disk part-way through
discarded *every* entry, not just one race's worth. Not in the original report, found
while reading, and the more damaging of the two.

Fix:

- Exclusive `flock` on a sidecar file across the whole read-modify-write. A sidecar
  rather than the index, because the index is replaced and a lock on a replaced inode
  protects nothing after the rename. Never unlinked (unlink race).
- The index is written to a uniquely-named sibling and `os.replace`d into position, so a
  failed write leaves the previous index intact.
- The temporary is opened `O_CREAT|O_EXCL` with `0o666` so the kernel applies the umask
  and default ACLs as `open(..., "w")` would; an existing index's mode is preserved.
  `tempfile.mkstemp` would force `0600` and lock other accounts out of a shared metadata
  directory.
- `flock` needs a writable descriptor, so a sidecar tighter than the index blocks a
  legitimate writer at the gate. It is **widened** to match the index when too tight —
  never narrowed, since an operator may deliberately open it to an account that takes
  locks without writing the index — and a writer that still cannot open it gets an error
  naming the sidecar and the repair instead of a bare `EACCES`.
- A filesystem that cannot do advisory locks (`EOPNOTSUPP`/`ENOLCK`/`ENOSYS`) falls back
  to the previous unlocked behaviour with a one-time warning, rather than failing a write
  that would otherwise succeed. A *refused* lock (`EACCES`) still raises.

The guarantee is exactly as strong as the filesystem's `flock`, stated in the docstring:
it holds locally and on shared filesystems honouring POSIX advisory locks across hosts,
and not where NFS or a cluster filesystem does not.

## ✨ Type of Change

- [ ] 💥 Breaking change (fix or feature that would cause existing functionality to not
      work as expected)
- [ ] 📝 Documentation update

- [x] **Unit Tests:** `uv run pytest tests/cli/test_signal_index_concurrency.py` — 8 tests
      covering both defects, the permission surfaces, the unsupported-filesystem fallback
      and descriptor hygiene.
- [x] **Full suite:** `uv run pytest` — 1060 passed, 23 skipped.
- [x] **Written test-first and confirmed failing on unfixed `main`,** with the specific
      failure modes rather than generic errors: the race leaves `{'11'}` missing `'22'`;
      the partial write leaves `partial: [` → `ParserError`.
- [x] **Discrimination measured, not assumed:** the race test passes **15/15 on the fix
      and 0/15 on unfixed code**. It refuses to pass a run the *scheduler* serialised
      rather than the lock — each writer records whether the two met inside the critical
      section and how long one waited for `LOCK_EX` (≈2.0 s under the fix, a 4× margin
      over the 0.5 s threshold) — retries three times, and fails loudly if no attempt
      raced.
- [x] **Mutation-tested:** eight targeted mutations, each killed by the test that owns
      it — `flock`→no-op, `LOCK_EX`→`LOCK_SH`, atomic replace→in-place write, drop the
      mode-preserving `chmod`, temp `0600`, never preserve an existing mode, drop sidecar
      alignment, sidecar→`0600`.

## 🏗️ Checklist

- [x] I have made corresponding changes to the documentation — the `update_signal_index`
      docstring stated the opposite ("Not safe against concurrent writers") and now
      states the guarantee and its filesystem caveat.

Not applicable.

## Not addressed, deliberately

- **No parent-directory `fsync`.** Durability, not atomicity: power loss can lose the
  last rename but cannot expose a partial index. Nothing else in the codebase fsyncs.
- **`LOCK_EX` blocks without a timeout.** `flock` releases on process death, so only a
  hung live writer blocks; `LOCK_NB`+retry turns a wait into a hard failure, worse for a
  long batch run.
- **No Windows locking.** Absent `fcntl`, updates run unlocked with a one-time warning
  rather than failing at import; the CI matrix is Linux and macOS.
- **Not verified across hosts.** All testing is single-host. Whether the cluster
  filesystems this project uses honour `flock` between machines is untested, and the
  fallback above is silent apart from one warning.

### DIFF
--- a/src/gwmock/cli/simulate_utils.py
+++ b/src/gwmock/cli/simulate_utils.py
@@ -8,13 +8,16 @@ import atexit
 import copy
 import json
 import logging
+import os
 import platform
 import re
 import shutil
 import signal
 import subprocess
+import tempfile
 import time
-from collections.abc import Callable
+from collections.abc import Callable, Iterator
+from contextlib import contextmanager
 from importlib import import_module
 from importlib import metadata as importlib_metadata
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


isaac-cf-wong:
@coderabbitai review

PR #325, opened 2026-08-07
docs(index): the concurrency guarantee does not hold across hosts

isaac-cf-wong:
## 📝 Summary

Tier: T1 (docs-only — no code semantics change; every changed line is prose inside a
docstring).

#323 stated that `update_signal_index` is "Safe against concurrent writers", qualified to
filesystems "that honour POSIX advisory locks across the participating hosts". Measured on an
NFS home shared by two machines: **that filesystem does honour them, and the guarantee still
fails.**

Two hosts, one metadata directory. The writer on host A committed event 101 and read it back.
The writer on host B **blocked 26.7 s** on the lock — exclusion works, measured twice at 28.8 s
and 26.7 s — then wrote an index containing only event 202. Event 101 was silently gone.

The cause is not the lock. Acquiring it revalidates the **sidecar**, not `signal_index.yaml`, so
host B read its client's cached view of the index, found nothing, started from an empty mapping,
and discarded A's entries while correctly holding the lock. A's data was on the server
throughout — the same process saw it seconds later once it opened the index directly.

It is structural rather than an oversight: `_atomically_write_index` replaces the index by
rename, which gives it a new inode, and a client holding a cached entry for the old path will not
resolve to it. The sidecar was chosen precisely so the lock survives that rename — and it is
therefore the file the kernel revalidates instead of the index.

This PR changes no code. It replaces a promise the code does not keep with the measured
limitation, and says plainly not to run concurrent writers against one metadata directory from
more than one host until the repair lands. Single-host concurrency is unaffected and remains
safe: one host has one cache.

Repair is designed separately; the leading candidate is a sequence number kept in the sidecar,
since the sidecar is the one file the lock does make coherent.

## ✨ Type of Change

      expected)
- [x] 📝 Documentation update

- [x] **Full suite:** `uv run pytest` on the review machine — 1060 passed, 23 skipped, unchanged
      from `main`.
- [x] **Two-host measurement** that produced the claim being corrected: writer on `alma10`,
      reader on `stadius-nc-5`, shared NFS home, with a four-snapshot cache diagnostic showing
      the index invisible while the lock was held and visible moments later once opened directly.
- [x] **Diff audited as prose-only** — no executable line changed.

## 🏗️ Checklist

- [x] I have made corresponding changes to the documentation — this PR *is* that change.
- [ ] I have added tests that prove my fix is effective — n/a, no behaviour change. The
      cross-host regression test belongs with the repair, and must be cross-host: a single-host
      test provably cannot catch this.

Not applicable.

## Summary by CodeRabbit

* **Documentation**
  * Clarified concurrency guarantees for signal index updates on single-host and multi-host filesystems.
  * Documented limitations that may cause lost updates when multiple hosts write concurrently.
  * Added guidance to avoid concurrent cross-host writers until the underlying issue is resolved.

### DIFF
--- a/src/gwmock/cli/simulate_utils.py
+++ b/src/gwmock/cli/simulate_utils.py
@@ -808,7 +808,7 @@ def update_signal_index(
     frame file(s) that contain it plus the batch metadata file, enabling O(1)
     signal->frame lookup by id.
 
-    **Safe against concurrent writers on one host**, which it was not before: the
+    **Safe against concurrent writers on one host, when the lock is taken**, which it was not before: the
     read-modify-write was unlocked, so two runs sharing a metadata directory lost one side's
     events -- reproduced deterministically with two processes and a barrier, where the loser's
     signals stayed in the frames while vanishing from the id lookup. The whole cycle now runs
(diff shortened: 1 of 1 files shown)


isaac-cf-wong:
@coderabbitai review

PR #327, opened 2026-08-07
fix(index): refuse to write the signal index on a stale read

isaac-cf-wong:
## 📝 Summary

#323 serialised `update_signal_index` with an exclusive `flock` on a sidecar. Measured
afterwards on a shared NFS home, two hosts, one metadata directory: **the lock worked and the
update was still lost.** The second writer blocked 26.7 s, then wrote an index holding only its
own event.

Acquiring the lock revalidates the **sidecar**, not `signal_index.yaml`. The reader had a cached
negative dentry for the index, saw nothing, started from `{}`, and discarded the first writer's
entries under a correctly held lock. Structural rather than an oversight:
`_atomically_write_index` replaces the index by rename — the new inode a cached dentry will not
resolve to — and the sidecar was chosen precisely so the lock survives that rename.

**The repair detects; it does not prevent.** The sidecar records a sha256 of the index bytes it
corresponds to, and a writer whose read disagrees raises `StaleIndexReadError` instead of
writing. Two clients stale in the same way would still agree with each other. The claim is that
silent loss becomes a loud refusal, nothing more.

Also here, each forced by review:

- the digest is computed from the bytes serialised, never a re-read — re-reading runs through
  the very path that can be stale, so a stale client would have recorded *its own view* as
  authoritative and certified the next stale write;
- the temporary is written binary, so the hashed bytes are the stored bytes (a text handle
  translates newlines on Windows and the digest could never match again);
- the no-op decision happens inside the lock — the sidecar is created by the same first write as
  the index, so an early prober caches a negative entry for **both**;
- an unreadable sidecar refuses rather than taking the permissive "predates the guard" path;
- a failure to record the digest raises `IndexDigestNotRecordedError` instead of being swallowed:
  the index is committed and correct, but the sidecar is behind, and every later write then
  refuses until it is re-baselined;
- neither error is retried — a backoff cannot fix a cache state, and they are raised after the
  frames are written, so each retry re-simulated the batch and tripped over its own outputs.

## ✨ Type of Change

      expected)
- [ ] 📝 Documentation update

- [x] **Unit tests:** `uv run pytest tests/cli/test_signal_index_stale_read.py` — 13 tests
      covering mismatch, match, no-digest, missing index, digest-vs-disk, unreadable sidecar, the
      unrecorded-digest raise, both-dentries-stale, and the retry rules.
- [x] **Full suite:** 1074 passed, 23 skipped.
- [x] **Mutation-tested**, each killed by the test that owns it: guard never raises; digest from
      a re-read; `exists()`-only early return; serialise after creating the temp; drop the
      mode-preserving chmod; and the no-retry rule in all three directions (never fires, always
      fires, follows `__context__` again).
- [x] **Cross-host, on the real fault** — writer `alma10`, reader `stadius-nc-5`, shared NFS,
      three fresh rounds with a settle before reading the truth: the fault reproduced in 2 of 3,
      the guard fired both times, **event 101 survived**, and the recorded digest matched a fresh
      hash of the index on disk in all three rounds. Before this change the identical scenario
      ended with event 101 gone.

## 🏗️ Checklist

- [x] I have made corresponding changes to the documentation — the docstring's guarantee states
      what holds and what does not.

Not applicable.

## Known limits, stated deliberately

- **Detection, not prevention** — established by review against this branch's own cross-host
  data: the guard fired, so it observed the stale view rather than revalidating past it.
- **A sidecar with no digest is accepted with a warning**, so the original bug is reachable for
  one write per unrecorded writer on a pre-upgrade directory. Same trade as #321's None-hash; it
  should become a refusal once no index predates this release.
- **The sibling `index.yaml` still has all three original defects** — unlocked read-modify-write,
  in-place truncating write, unparsable→new. Reproduced at 4/20 concurrent trials, and invisible
  today only because nothing in production reads it. Tracked separately; not folded in here,
  since one work item is one PR.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved signal-index update reliability by preventing stale or conflicting changes from overwriting newer data.
  * Preserved the last valid, readable index when an update fails during publication.
  * Added safer handling for empty and withdrawal-only updates.
  * Improved retry behavior so non-recoverable failures are not retried unnecessarily.
  * Added compatibility support and warnings for older index metadata formats.

### DIFF
--- a/src/gwmock/cli/simulate_utils.py
+++ b/src/gwmock/cli/simulate_utils.py
@@ -494,6 +494,29 @@ class IndexDigestNotRecordedError(RuntimeError):
 _NOT_WORTH_RETRYING = (StaleIndexReadError, IndexDigestNotRecordedError)
 
 
+def _is_not_worth_retrying(error: BaseException) -> bool:
+    """Return whether this failure, or anything it was raised from, is futile to retry.
+
+    The chain is walked because matching the outer type alone is fragile: wrapping one of these
+    in anything else -- ``raise RuntimeError(...) from StaleIndexReadError`` -- would restore the
+    retries, and a stale index read is no more fixable by waiting for having been re-raised.
+
+    Args:
+        error: The exception that ended the attempt.
+
+    Returns:
+        Whether to give up immediately.
+    """
+    seen: set[int] = set()
+    current: BaseException | None = error
+    while current is not None and id(current) not in seen:
+        if isinstance(current, _NOT_WORTH_RETRYING):
+            return True
+        seen.add(id(current))
+        current = current.__cause__ or current.__context__
+    return False
+
+
 def retry_with_backoff(
     func: Callable[..., Any],
     max_retries: int = 3,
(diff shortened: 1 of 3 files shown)


isaac-cf-wong:
@coderabbitai review

PR #329, opened 2026-08-07
fix(index): lock and atomically write the metadata index

isaac-cf-wong:
## 📝 Summary

`update_metadata_index` writes `index.yaml` with two of the three faults `signal_index.yaml` had
before #323, from the same caller (`save_batch_metadata`) into the same metadata directory:

- an **unlocked read-modify-write** — two runs sharing a directory lose one side's entries.
  Measured during review at **4 losses in 20 concurrent trials**;
- an **in-place truncating write**, whose failure the loader above then reads as "create a new
  index" — so one interrupted dump discards *every* entry, not one.

Found at the end of the signal-index repair by asking what several rounds of review had missed:
nobody had looked at the file written on the line above.

The fix reuses the helpers the signal index already has — both take the index path — so this is
wiring rather than new invention: the update runs under the same exclusive `flock` on a sidecar,
and the result is renamed into place instead of truncating.

**No staleness digest here, deliberately.** That guard's job is to refuse a write, and nothing in
production reads `index.yaml` today, so aborting a run over it would cost more than it protects.
The residue is stated precisely in the docstring, because it is worse than it first sounds: a
stale negative dentry makes the loader start from an empty mapping, so a cross-host stale write
replaces the **whole index**, not one row. Measured: `{data-0, data-1}` became `{data-2}`. The
condition to revisit is therefore a **reader-writer from another process**, not merely a reader.

## ✨ Type of Change

      expected)
- [ ] 📝 Documentation update

- [x] **Unit tests:** `uv run pytest tests/cli/test_metadata_index_durability.py` — 3 tests. Two
      were written first and **confirmed failing against the unfixed implementation**; the third
      guards mode preservation.
- [x] **Full suite:** 1078 passed, 23 skipped.
- [x] **Mutation-tested**, each killed by the test that owns it and nothing else: drop the lock →
      the race test; write in place → the failed-write test.
- [x] The race test carries the validity check from the signal-index work: it refuses to pass a
      run the *scheduler* serialised rather than the lock. That check earned its place
      immediately — an earlier version armed its rendezvous on `yaml.safe_load`, which only fires
      when the index already exists, so against a fresh directory nothing overlapped. It reported
      "scenario did not run" instead of passing green. The index is now seeded first.

## 🏗️ Checklist

- [x] I have made corresponding changes to the documentation — the docstring states the
      guarantee, the residue and the revisit condition.

Not applicable.

## Known limits

- **Two files in one directory now carry different guarantees.** Deliberate, reviewed, and
  recorded in the docstring rather than only in this description.
- **The permissions test does not discriminate this change** — it passes on the in-place
  implementation for the same reason, and guards the shared writer's mode behaviour. Said so in
  its docstring so nobody mistakes it for coverage of atomic-vs-in-place.
- **`save_batch_metadata` now takes two locks per batch**, sequentially and never nested;
  measured at ~3.3 ms per update, dominated by the atomic write's `fsync` rather than the lock.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved metadata index updates to safely handle concurrent changes.
  * Prevented interrupted or failed writes from corrupting or replacing existing index data.
  * Preserved file permissions when updating the metadata index.

* **Tests**
  * Added coverage for concurrent updates, failed writes, temporary-file cleanup, and permission preservation.

### DIFF
--- a/src/gwmock/cli/simulate_utils.py
+++ b/src/gwmock/cli/simulate_utils.py
@@ -598,6 +598,20 @@ def update_metadata_index(
     The index maps data file names to their corresponding metadata files,
     enabling O(1) lookup to find metadata for a given data file.
 
+    Serialised and written atomically, for the reasons its sibling
+    :func:`update_signal_index` was: an unlocked read-modify-write let two runs sharing a
+    metadata directory lose one side's entries (measured at 4 losses in 20 concurrent trials),
+    and the in-place truncating write turned one interrupted dump into the loss of *every* entry,
+    because the loader above treats an unparsable index as "create a new index".
+
+    **Deliberately weaker than :func:`update_signal_index`: there is no staleness digest here.**
+    That guard's job is to refuse a write, and on a shared filesystem a client can hold a stale
+    view of this file just as it can of the signal index. Nothing in production reads
+    ``index.yaml`` today, so aborting a run over it would cost more than the entry it protects.
+    The residue is that a cross-host stale read can still drop an entry here silently. **Revisit
+    when something starts reading it** -- the guarantee this file offers is then the guarantee
+    that consumer inherits.
+
     Args:
         metadata_directory: Directory where metadata files are stored
         output_files: List of output data file Paths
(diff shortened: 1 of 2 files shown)


isaac-cf-wong:
@coderabbitai review

PR #332, opened 2026-08-08
fix(placement): stop claiming events whose waveform ended before the segment

isaac-cf-wong:
## 📝 Summary

`_event_starts_before_segment_end` had no lower bound. In a sequential run that is invisible,
because earlier segments have already consumed the earlier events. It is not invisible in a run
whose `start-time` is later than its population's first event: nothing has consumed those events,
every one of them satisfies "starts before this segment ends", and the first segment claims the
entire back catalogue.

**Measured on the shipped ET config:** a run starting 53 ks into the catalogue batched **126
events for a segment containing 13**, estimated 12.5 GiB, and was refused by the preflight; at
41 ks it batched 98 for 16; at 4 ks, 19 for 8. The batch tracks the distance from the catalogue
start rather than the local event density, and the extra events' samples are then cropped as
out-of-segment — generated and discarded. Resuming a campaign mid-catalogue, or slicing a long
population into per-day jobs, is exactly this.

This adds the missing bound, using `post_coalescence_duration` from gwmock-signal 0.16.0:

- `AdapterOrchestrator._event_ended_before_segment_start` — `coa_time + tail < segment_start`,
  strict, so an event ending exactly on the boundary keeps its final sample.
- `SignalAdapter.post_coalescence_duration` — the wrapper the orchestrator actually talks to,
  mirroring `pre_coalescence_duration` including its unsupported-parameter retry.
- Both consumption loops **step over and consume** a finished event.
- Dependency floor moved to `>=0.16.0` in all four places, with `uv.lock`. The two comments
  justifying the old floor had become false and are rewritten.

**`None` means unknown and claims the event.** The tail is a fixed fraction of the buffer rather
than a physical ringdown — 0.4 s for a 4 s BBH buffer, 25.6 s for a 256 s BNS one — so an event
coalescing before the segment may still be producing signal inside it. Reading unknown as zero
would discard those silently, and for a whole backend at a time.

### Two design points worth stating

**The bound is not another clause in the existing predicate.** Both loops `break` on the first
event that does not belong, and `population_index` means "every event before this one is consumed".
An early event answering *does not belong* would stop the walk at position 0 and repeat that on
every following segment — the run would **stall permanently**, which is worse than the
over-batching being fixed.

**The index advances by the walk's traversal, not by the skip count.** Skips are not a prefix:
consumption is ordered by waveform *start* while skipping is decided by waveform *end*, so a long
overlapping event can be followed by a short finished one. Advancing by skips leaves the claimed
event behind the index, and a direct library caller retrying after a generation failure would never
generate it. Only the all-skipped batch advances inside the walk, because the caller returns early
on an empty batch and never reaches the commit.

## ✨ Type of Change

      to not work as expected)
- [ ] 📝 Documentation update

- [x] **Unit Tests:** `uv run pytest` → **1099 passed, 23 skipped** on the review machine.
- [x] **Regression first.** Five tests were run against the unfixed tree and confirmed failing
      there *for the right reason* — 13 batched where 3 belong, and 12/22/42 growing with the
      offset, reproducing the measured shape rather than erroring.
- [x] **Checked against generation, not against another prediction.** The predicted tail is
      compared with where the produced `TimeSeries` actually ends, and lead + tail against its
      duration. Inflating the tail by 5% fails both and nothing else.
- [x] **Three tests use the real adapter** with nothing substituted — see below for why that
      matters more than the count suggests.
- [x] **Fifteen mutations across three rounds, all killed**, each by the test that owns it:
      unknown-as-zero, skip-as-break, index-not-advanced, `<`→`<=`, per-event loop left one-sided,
      no-`coa_time`-as-finished, adapter method removed, adapter answering `None`,
      commit-by-generated-count, walk-advancing-by-skips, guard restored, post-side retry removed,
      query bypassing the parameter merge, tail inflated 5%.

### What review caught that the tests did not

**The first version of this change was a no-op in production while fifteen tests passed.** Both
independent reviewers found it in the same round. `signal_adapter` is this package's own
`SignalAdapter`, not gwmock-signal's `CBCSimulator`, and it had only `pre_coalescence_duration`; a
`getattr` guard turned every tail into "unknown" and the bound never fired. Every test supplied a
stub built to answer, so the stub *was* the bug's disguise. The guard is gone — the query is called
directly, so a missing method warns once through the existing handler instead of vanishing, which
is this codebase's own rule about fallbacks having reported costs.

Round 2 found no logic defects and closed coverage gaps instead: the query-versus-generation
agreement test above, the post-side retry, the parameter merge, and warn-once dedup across a walk.

## 🏗️ Checklist

      works.

**On the unchecked box, and one gap left open deliberately.**

No user-guide coverage for the tail. A reviewer checked and the *pre* side has none either, so this
is symmetric with existing precedent rather than a new hole; a short note is warranted and is
tracked separately.

**The ripple batched path never exercises a skip end to end.** The real-adapter tests use the
default LAL backend and the bundled population has its one event inside the segment, so index
agreement between the two execution modes is validated with *zero skips*. Closing it needs the
`[jax]` extra and belongs in the `test-jax` cells; it is tracked. Related, pre-existing, and shared
with the pre side: in a batched config with no `waveform-backend` key, the adapter holds LAL while
`simulate_cbc_batch` generates with ripple (4.0 s vs 3.125 s buffers for the same 30+25 binary).
LAL's tail is the larger, so it errs toward claiming — wasted work, not dropped signal — and the
shipped batched example sets ripple explicitly. Safe today, but by asymmetry rather than by design.

## Summary by CodeRabbit

* **New Features**
  * Improved event placement by skipping waveforms that finish before a segment begins.
  * Added waveform tail-duration detection to support more accurate event processing.
  * Added fallback handling when tail duration cannot be determined, with deduplicated warnings.

* **Bug Fixes**
  * Preserved catalogue indexing and traversal across skipped events, failed generations, and empty batches.
  * Improved consistency between batched and per-event processing.
  * Updated the minimum supported signal package version.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -22,18 +22,23 @@ classifiers = [
 requires-python = ">=3.12"
 dynamic = ["version"]
 dependencies = [
-  # 0.15.0 is the first release exposing `CBCSimulator.pre_coalescence_duration`, which segment
-  # placement asks before generating so an event is claimed by the segment its *waveform* starts in
-  # rather than the one holding its coalescence. Below this the query is absent, placement falls back
-  # to `coa_time`, and the start of any waveform landing near a boundary is cropped -- so this is a
-  # floor the behaviour depends on, not a convenience.
+  # Two floors, both behavioural rather than conveniences.
   #
-  # Floored at 0.16.0 for a *correctness* fix rather than a feature: gwmock-signal #370 projects the
-  # source direction into the frame of date. Before it, `gha = GMST - right_ascension` mixed a J2000
-  # right ascension with an Earth rotation measured from the mean equinox of date, an inconsistency
-  # worth 0.43 degrees in right ascension by 2030. The fix cuts worst-case disagreement with
-  # `lalpulsar.Barycenter` from 1.774e-04 s to 8.686e-07 s across a 192-point sky/epoch grid. It
-  # changes continuous-wave output, which is why this bump carries regenerated e2e references.
+  # 0.15.0 first exposed `CBCSimulator.pre_coalescence_duration`, which segment placement asks
+  # before generating so an event is claimed by the segment its *waveform* starts in rather than the
+  # one holding its coalescence. Below it the query is absent, placement falls back to `coa_time`,
+  # and the start of any waveform landing near a boundary is cropped.
+  #
+  # 0.16.0 adds the other half, `post_coalescence_duration`. Without the tail, a run whose
+  # start-time is later than its population's first event cannot tell that the earlier events are
+  # finished, so it claims every one of them into its first segment and discards the samples; the
+  # batch then grows with the distance from the catalogue start until the preflight refuses it.
+  #
+  # 0.16.0 also carries a projection *correctness* fix (gwmock-signal #370): the source direction is
+  # projected into the frame of date, where before `gha = GMST - right_ascension` mixed a J2000 right
+  # ascension with an Earth rotation measured from the mean equinox of date -- 0.43 degrees in right
+  # ascension by 2030. It changes continuous-wave output, which is why the bump landed with
+  # regenerated e2e references rather than inside this change.
   "gwmock-signal>=0.16.0",
   "gwmock-noise>=0.10.1",
   "gwmock-pop>=0.11.5",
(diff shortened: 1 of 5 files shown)


isaac-cf-wong:
@coderabbitai review

PR #333, opened 2026-08-08
chore(deps): floor gwmock-signal at 0.16.0 for the frame-of-date projection fix

isaac-cf-wong:
## 📝 Summary

Floors `gwmock-signal` at 0.16.0. **Not a routine bump** — 0.16.0 contains a projection
correctness fix that changes continuous-wave output, so it is split out from the placement work
that needed the floor (#332) rather than landing inside it. A later bisect of CW values should
arrive at this commit, not at a PR titled as a placement fix.

**The fix.** gwmock-signal
[#370](https://github.com/Leuven-Gravity-Institute/gwmock-signal/pull/370) projects the source
direction into the frame of date. Before it, the projection computed
`gha = GMST - right_ascension` using a **J2000** right ascension, while GMST measures the Earth's
rotation from the mean equinox **of date**. The two live in different frames, and the mismatch is
the precession accumulated since J2000 — 0.43° in right ascension by 2030. An inconsistency, not a
convention.

**Anchored outside gwmock-signal's own arithmetic.** The fix cuts worst-case disagreement with
`lalpulsar.Barycenter` from **1.774e-04 s to 8.686e-07 s** across a 192-point grid (3 detectors ×
4 right ascensions × 4 declinations × 4 epochs). The residue is nutation-scale — which LAL adds
separately and gwmock-signal does not model — so 8.7e-07 s is the size the omission predicts rather
than an unexplained leftover.

Both comments justifying the previous floor named 0.15.0 and are corrected.

## ✨ Type of Change

      to not work as expected)
- [ ] 📝 Documentation update

*(Closest available box. Strictly this is a dependency floor whose observable effect is a
corrected continuous-wave projection — no gwmock API changes, but the output does change; the
regenerated references below are that change made explicit.)*

- [x] **e2e:** `uv run pytest -m e2e -k cw` → 6 passed against the regenerated references.
- [x] **Generated on the platform CI compares against** — Linux x86_64, Python 3.12 — not on the
      arm64 review machine. `argmax` is compared **exactly**, and a CW signal is near-sinusoidal, so
      its peak hops between nearly-equal maxima under a tiny amplitude change; references written
      elsewhere could fail on `argmax` alone.
- [x] **The two `argmax` moves recorded here (653 → 1165, 1644 → 1849) are identical to the ones CI
      observed**, and the regenerated peaks agree with CI's values to ~7 digits.
- [x] **Validated before regenerating, not after.** The CW check was confirmed *failing* on the
      generating host first. Had it passed there, the drift would have been CI-specific and
      regenerating from that host would have blessed the wrong numbers.
- [x] **Scope:** the file set, the fingerprint (`{Linux, x86_64, 3.12}`) and every other scenario's
      reference are unchanged; one file differs.
- [x] **Attribution proven, not assumed.** `main` with only these four pins bumped and no other
      change reproduces the same drift to ~13 significant digits, which is what established that
      #332's placement work was not responsible.

### Magnitude, and what is not anchored

The shifts are **peak 2.4e-04 to 7.9e-03** and **rms 2.2e-03 to 6.2e-03** relative. The correction
is 0.43° = 7.5e-03 rad and antenna patterns vary on radian scales, so a fractional amplitude change
of that order is what it predicts, with the smaller values where the pattern derivative is flatter.

**Not anchored:** where the new `argmax` lands. It is consistent with peak-hopping on a
near-sinusoid, but nothing here predicts the specific sample.

## 🏗️ Checklist

      works.

**On the unchecked box:** no documentation change. The corrected projection is described in
gwmock-signal's own release, and this repository's docs do not state the old convention anywhere I
could find — but I did not exhaustively verify that, so it is worth a reviewer's grep rather than my
assurance.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -27,7 +27,14 @@ dependencies = [
   # rather than the one holding its coalescence. Below this the query is absent, placement falls back
   # to `coa_time`, and the start of any waveform landing near a boundary is cropped -- so this is a
   # floor the behaviour depends on, not a convenience.
-  "gwmock-signal>=0.15.0",
+  #
+  # Floored at 0.16.0 for a *correctness* fix rather than a feature: gwmock-signal #370 projects the
+  # source direction into the frame of date. Before it, `gha = GMST - right_ascension` mixed a J2000
+  # right ascension with an Earth rotation measured from the mean equinox of date, an inconsistency
+  # worth 0.43 degrees in right ascension by 2030. The fix cuts worst-case disagreement with
+  # `lalpulsar.Barycenter` from 1.774e-04 s to 8.686e-07 s across a 192-point sky/epoch grid. It
+  # changes continuous-wave output, which is why this bump carries regenerated e2e references.
+  "gwmock-signal>=0.16.0",
   "gwmock-noise>=0.10.1",
   "gwmock-pop>=0.11.5",
   "typer>=0.19.0",         # Literal-type CLI options; the SPEC 0 floor (0.13) predates them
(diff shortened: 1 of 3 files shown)


isaac-cf-wong:
@coderabbitai review

PR #334, opened 2026-08-08
fix(placement): answer placement with the backend that generates the buffer

isaac-cf-wong:
## 📝 Summary

`execution: batched` always generates with ripple, but a batched config naming no
`waveform-backend` built its adapter on LAL. The placement queries deciding where each waveform
starts and ends therefore described a **different library's buffer** than the one generated.

The two libraries round buffer lengths onto different grids, so which tail is longer flips with
mass and cutoff. Across 24 mass/cutoff combinations, **9 have ripple's tail longer than LAL's**:

| binary | cutoff | LAL answered | ripple generates | content droppable |
| --- | --- | --- | --- | --- |
| 1.4+1.35 | 5 Hz | 819.20 s | 934.17 s | **115.0 s** |
| 3.0+3.0 | 5 Hz | 204.80 s | 256.00 s | **51.2 s** |
| 4.5+3.0 | 10 Hz | 25.60 s | 29.66 s | 4.1 s |
| 1.4+1.35 | 30 Hz | 6.40 s | 8.44 s | 2.0 s |

On those combinations an event whose ripple content reaches into the segment is reported finished
on LAL's shorter answer and skipped, deleting signal — the outcome the unknown-is-not-zero rule in
#332 exists to prevent. Low mass at low cutoff is the ET regime.

### What changed

- **A batched config naming no `waveform-backend` now builds its adapter on ripple**, so placement
  and generation describe one buffer. This aligns `pre_coalescence_duration` as well as
  `post_coalescence_duration`; both had the mismatch.
- **When ripple is not installed the tail reports unknown instead of answering.** Unknown claims the
  event, so nothing is skipped on a number this run cannot stand behind; the batch stays non-empty
  and generation raises ripple's own install error. Answering with the default library could skip
  events, and because the batched path returns early on an empty batch, generation would never run
  and the install error would never surface.
- **Provenance names the library that ran.** A substituted backend is recorded where the
  substitution happens, and the record is cleared when `signal_adapter` is replaced, so it cannot
  outlive the adapter it describes.

### ⚠️ Behaviour change

**Boundary events move** for batched configs naming no `waveform-backend`. LAL and ripple also
disagree on the lead (30+25 at 30 Hz: 3.60 s against 2.81 s), so an event at `end + 3.0` was claimed
by this segment before and belongs to the next one after. The new placement is the correct one, but
a run's segmentation can change. Approved by the maintainer.

The shipped e2e scenario is unaffected — its output is byte-identical before and after, because it
has no event near a boundary.

## ✨ Type of Change

- [x] 💥 Breaking change (fix or feature that would cause existing functionality
      to not work as expected)
- [ ] 📝 Documentation update

*(Both: it fixes data loss, and it changes which segment claims a boundary event.)*

- [x] **Unit Tests:** `uv run pytest` → **1116 passed, 23 skipped**.
- [x] **Equality sweeps rather than a single point:** six mass/cutoff points for the tail and three
      for the lead assert the query equals the generating backend. All six tail points fail with the
      fix reverted, including the one that passed beforehand — equality is stronger than the
      inequality it replaces.
- [x] **The skip is exercised end to end through real ripple**, not a stub, and both execution modes
      are checked to agree *with skips actually taken*.
- [x] **Ten mutations, each killed by the test that owns it**, including one reinstating the
      original defect, one restoring the silent fallback, and one leaking the substitution record
      into provenance.
- [x] **Without the `[jax]` extra:** construction still succeeds, the tail reports unknown, and
      nothing is skipped.
- [x] **e2e compared by content against unmodified `main`:** byte-identical.

## 🏗️ Checklist

      works.

**On the unchecked box:** no user-guide entry for the placement change. The pre side has no guide
coverage either, so this matches existing precedent rather than opening a new gap, and it is tracked
separately.

No UI. The output this changes is the placement answer, measured against the installed backends
before the fix:

```
 30.0+25.0  f= 30.0  LAL     0.400391  ripple     0.312500  safe=True
 30.0+25.0  f= 20.0  LAL     0.400391  ripple     0.450195  safe=False
 30.0+25.0  f= 10.0  LAL     1.599609  ripple     1.423828  safe=True
 30.0+25.0  f=  5.0  LAL     6.400391  ripple     7.119141  safe=False
 10.0+8.0   f= 30.0  LAL     0.799805  ripple     0.659180  safe=True
 10.0+8.0   f= 20.0  LAL     1.599609  ripple     1.406250  safe=True
 10.0+8.0   f= 10.0  LAL     6.400391  ripple     7.119141  safe=False
 10.0+8.0   f=  5.0  LAL    51.200195  ripple    42.714844  safe=True
  4.5+3.0   f= 30.0  LAL     1.599609  ripple     1.875000  safe=False
  4.5+3.0   f= 20.0  LAL     6.400391  ripple     5.000000  safe=True
  4.5+3.0   f= 10.0  LAL    25.599609  ripple    29.663086  safe=False
  4.5+3.0   f=  5.0  LAL   204.799805  ripple   184.528320  safe=True
  3.0+3.0   f= 30.0  LAL     3.200195  ripple     2.500000  safe=True
  3.0+3.0   f= 20.0  LAL     6.400391  ripple     6.750000  safe=False
  3.0+3.0   f= 10.0  LAL    51.200195  ripple    41.005859  safe=True
  3.0+3.0   f=  5.0  LAL   204.799805  ripple   256.000000  safe=False
  1.4+1.35  f= 30.0  LAL     6.400391  ripple     8.437500  safe=False
  1.4+1.35  f= 20.0  LAL    25.599609  ripple    23.730469  safe=True
  1.4+1.35  f= 10.0  LAL   204.799805  ripple   148.315430  safe=True
  1.4+1.35  f=  5.0  LAL   819.200195  ripple   934.173828  safe=False
 60.0+50.0  f= 30.0  LAL     0.400391  ripple     0.263672  safe=True
 60.0+50.0  f= 20.0  LAL     0.400391  ripple     0.312500  safe=True
 60.0+50.0  f= 10.0  LAL     0.799805  ripple     0.674805  safe=True
 60.0+50.0  f=  5.0  LAL     3.200195  ripple     2.636719  safe=True

UNSAFE combinations: 9 of 24
```

`safe=False` marks a row where the query would report an event finished while ripple is still
producing it. After the fix both columns are the same number by construction, which is what the
equality sweep asserts.

### DIFF
--- a/src/gwmock/cli/adapter_orchestration.py
+++ b/src/gwmock/cli/adapter_orchestration.py
@@ -805,28 +805,25 @@ class AdapterOrchestrator(TimeSeriesMixin, Simulator):
         if getattr(self, "_substituted_waveform_backend", None) is not None:
             self._substituted_waveform_backend = None
 
-    def _require_placement_backend(self) -> None:
-        """Refuse to answer a placement question the run cannot answer correctly.
+    def _placement_backend_can_answer(self) -> bool:
+        """Whether the tail query describes the buffer this run will actually generate.
 
         A batched config with no ``waveform-backend`` needs ripple, because that is what it
-        generates with. When ripple is missing the adapter is left on LAL, and a LAL answer here is
-        not merely unavailable -- it is *wrong in a way that consumes events*: it can report an event
-        finished when ripple would still be producing it, and the batched path then returns early on
-        the empty batch, so generation never runs and the install error never surfaces. The event is
-        skipped, the index advances, and nothing is reported.
+        generates with. When ripple is missing the adapter is left on the default library, whose
+        tail is shorter than ripple's at some masses and cutoffs -- so *acting* on that answer can
+        skip an event ripple would still be producing.
 
-        Raises:
-            ImportError: If placement is asked of a batched run whose generating library is absent.
+        Only the skip is dangerous, so only the skip is withheld: the tail reports *unknown*, which
+        by this module's standing rule claims the event rather than dropping it. The batch is then
+        non-empty, generation runs, and ripple's own install error surfaces where it belongs.
+        Refusing to answer at all was tried and was too broad -- it made every batched unit test and
+        every construct-and-inspect caller raise, in an environment where nothing was going to
+        generate anyway.
+
+        Returns:
+            Whether the answering backend is the one that will generate.
         """
-        if self._substituted_waveform_backend != _RIPPLE_UNAVAILABLE:
-            return
-        raise ImportError(
-            "execution: batched generates with ripple, so segment placement has to ask ripple where "
-            "each waveform starts and ends -- and ripple is not installed. Answering with the "
-            "default library instead would skip events it reports as finished while ripple would "
-            "still be producing them, without ever reaching the code that raises this at "
-            "generation. Install the extra: pip install 'gwmock-signal[jax]'."
-        )
+        return self._substituted_waveform_backend != _RIPPLE_UNAVAILABLE
 
     def _event_ended_before_segment_start(self, parameters: Mapping[str, Any]) -> bool:
         """Return whether this event's waveform has finished before the current segment begins.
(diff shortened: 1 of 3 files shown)


isaac-cf-wong:
@coderabbitai review

PR #335, opened 2026-08-08
test(index): exercise the unlocked index branches, and remove their coverage pragma

isaac-cf-wong:
## 📝 Summary

The signal index has two branches that degrade to unsynchronised writes: no `fcntl` module
(Windows), and a filesystem that rejects `flock` (some network mounts). Both carried
`# pragma: no cover`, and the CI matrix is Linux and macOS, so nothing exercised either.

It showed. The `fcntl is None` branch broke **three times in one day**, each time while something
else was being fixed, each time caught by a reviewer reading rather than by a failure:

1. `os.fchmod` — Unix-only, so the second write failed.
2. the digest re-read.
3. `_record_digest` opening `"r+"` on a file that branch never creates, which failed **every** update
   after the index was committed, with a recovery message naming a file that did not exist.

### What changed

- **Nine tests** reach both degraded paths, pinning what a hand probe had checked by eye: both events
  written, the recorded digest matching the bytes on disk, and the unlocked warning arriving once —
  including once across *two directories*, which distinguishes once-per-process from
  once-per-whatever-the-test-used.
- **`# pragma: no cover` removed** from the `fcntl is None` branch, after removing it and
  re-measuring. Pragma'd lines are excluded from the report, so their absence from "missing" proves
  nothing.
- **No source behaviour changes.** One comment and the pragma are the only non-test edits.

### The fixture is a tripwire, not a platform simulation

Stated plainly because two review rounds caught this docstring claiming otherwise:

- Removing `fcntl` makes the branch reachable. That is all it does.
- `os.fchmod` and `os.fchown` are deleted as **tripwires** against those names returning to this
  path. They are currently *off-path* — breakage 1's own fix replaced `fchmod` with `os.chmod` — but
  they do catch breakage 1 if it returns, which is verified below.
- **Nothing here simulates Windows.** An attempt to add one was reverted: I had `os.chmod` reject a
  descriptor "as Windows does", but `PATH_HAVE_FCHMOD` is forced to 1 for `MS_WINDOWS` in CPython's
  `posixmodule.c`, so `os.chmod(fd, mode)` succeeds on Windows for every interpreter this project
  supports. That guard would have failed a refactor which is correct on the platform it claimed to
  model.

### One commit in this branch adds something and a later one takes it back

`811dc4b` added `O_BINARY` to the atomic write's `os.open`, on the reasoning that a Windows
descriptor would otherwise translate newlines beneath `fdopen(..., "wb")` and leave the digest
describing different bytes than the file. `bf48265` reverts it: `io.FileIO.__init__` runs
`_setmode(self->fd, O_BINARY)` under `#if defined(MS_WINDOWS) || defined(__CYGWIN__)`, and it does so
for a caller-supplied descriptor as well as one it opened. The failure cannot occur on any CPython,
the flag was redundant, and **the original code and comment were correct before being touched.**

Both commits are kept rather than rewritten, because the second is where the verification lives.

## ✨ Type of Change

      to not work as expected)
- [ ] 📝 Documentation update

*(Tests, one comment, and a coverage pragma. No behaviour changes.)*

- [x] **Unit Tests:** `uv run pytest` → **1125 passed, 23 skipped**.
- [x] **All three historical breakages reinstated and killed**, each by tests that own it — B1 by the
      `fchmod` tripwire, B2 (digest accumulating via `O_APPEND`) by 4 tests, B3 (`"r+"` with no create
      fallback) by 4 tests.
- [x] **The pragma removal was measured, not inferred:** removed, re-run, lines confirmed executed.
- [x] The mutation harness refuses to apply a pattern it cannot match, and verifies its restore
      byte-for-byte, so a mutation that fails to land is reported rather than counted as survival.

## 🏗️ Checklist

      works.

### Known limits of this coverage, so the report is not read as more than it is

- `fail_under = 0` and there is no Codecov gate, so removing the pragma changes the report's
  accounting, not enforcement. The accounting is now accurate.
- The branch is covered **by monkeypatched tests only**. No report distinguishes mocked coverage from
  real, and a behaviour-only regression on Windows would still read green — the same as under the
  pragma, which at least advertised ignorance.
- Two genuine Windows differences on this path are tracked separately rather than tested here: the
  index rename is not fsynced before its digest is recorded (loud false `StaleIndexReadError` after a
  crash, not corruption), and `os.replace` fails on Windows while any process — including a *reader*
  such as `find_signals` — holds the index open.

No UI.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of signal-index updates when file locking is unavailable.
  * Ensured sequential updates preserve all events and keep index digests accurate.
  * Ensured digest sidecar files are created and populated during unlocked updates.
  * Prevented repeated unlocked-mode warnings within the same process.

### DIFF
--- a/src/gwmock/cli/simulate_utils.py
+++ b/src/gwmock/cli/simulate_utils.py
@@ -728,9 +728,10 @@ def _exclusive_index_lock(index_file: Path) -> Iterator[None]:
     """
     # No `# pragma: no cover` any more, and that is the point: this branch broke three times in one
     # day while it was excluded, each time caught by a reviewer reading rather than by a failure.
-    # `tests/cli/test_signal_index_without_locking.py` reaches it by presenting the platform as
-    # Windows -- no `fcntl` *and* no Unix-only `os` calls, because nulling the module alone left
-    # `os.fchmod` working and one of those three defects therefore undetectable.
+    # `tests/cli/test_signal_index_without_locking.py` makes it reachable by setting `fcntl` to
+    # `None`, and tripwires `os.fchmod` and `os.fchown` so a return of those two names to this path
+    # fails a test. It does **not** simulate Windows -- nothing here does, and the real differences
+    # that remain are tracked rather than tested.
     if fcntl is None:
         _warn_unlocked_once()
         yield
--- a/src/gwmock/cli/simulate_utils.py
+++ b/src/gwmock/cli/simulate_utils.py
@@ -943,13 +943,7 @@ def _atomically_write_index(index_file: Path, index: dict[str, Any]) -> str:
     digest = hashlib.sha256(payload).hexdigest()
     existing_mode = _existing_index_mode(index_file)
     temporary = index_file.with_name(f"{index_file.name}.{uuid.uuid4().hex}.tmp")
-    # `O_BINARY` is what makes the comment below true rather than aspirational. Without it a
-    # descriptor opens in the CRT's *text* mode on Windows, and that translation happens beneath
-    # `os.fdopen(..., "wb")` -- so the bytes written would be CRLF while the digest describes LF,
-    # and every update after the first would refuse a perfectly good index as stale. It is 0 on
-    # POSIX, where the distinction does not exist, so `getattr` keeps one code path.
-    binary_flag = getattr(os, "O_BINARY", 0)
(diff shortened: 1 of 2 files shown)


isaac-cf-wong:
@coderabbitai review

PR #336, opened 2026-08-08
fix(cli): make the signal index's rename durable before its digest describes it

isaac-cf-wong:
## 📝 Summary

`_atomically_write_index` fsynced the temporary's *contents* and then `os.replace`d it, but never
flushed the parent directory. `os.replace` is atomic with respect to readers; atomicity is not
durability, so the directory entry could still be in the page cache when `_record_digest` wrote the
sidecar. A crash in that window reboots to the previous index while the sidecar describes the new one,
and every later write then refuses as stale against a perfectly good file — by the function's own
docstring, a permanent wedge until an operator deletes the sidecar.

The directory is now flushed between the rename and the digest write. Where that flush cannot be
performed, the digest is still recorded and the degradation is reported:

- **The filesystem refuses `fsync` on a directory** (some network mounts): warned once per filesystem,
  stating that no repair exists on that mount and naming the alternative.
- **The directory cannot be opened for reading** (its own permissions, e.g. write-only): warned once
  per directory, stating the actual repair — make it readable; deleting the sidecar does not help,
  because the next write cannot flush it either.
- **The platform has no directory-flush primitive** (Windows): records and stays silent. A warning on
  every run about a gap the operator cannot close only buries the ones that can be acted on.

Both degraded states clear when a flush there succeeds, so a condition that ends and returns is
reported again.

Recording the digest regardless is deliberate. Withholding it leaves an empty sidecar, and an empty
sidecar is the permissive legacy path: a writer on another host with a stale cached view is accepted
and discards the entries it could not see — silent loss, no crash needed, on exactly the mounts this
guard exists for. The exposure that remains is a crash inside the unflushed window, which produces a
refusal against a good index: loud, and recoverable. A rare loud failure is worth keeping over a
routine silent one. On a refusing NFS mount that window lasts until the client sends the rename —
seconds, not instants — so the trade does not rest on the window being small.

No Windows `FlushFileBuffers` path is included: no host in this project can execute it, and untested
platform code that looks like protection is worse than a documented gap.

Tier: **T2** (behaviour).

## ✨ Type of Change

- [ ] 📝 Documentation update

- [x] **Unit tests:** `uv run pytest` — **1140 passed** with the `jax` extra, **1098 passed** without it
      (CI's plain `uv sync --group dev` cell; `ripplegw` is the axis that differs, since `jax` itself is
      a base dependency of gwmock-pop).
- [x] **New tests:** `tests/cli/test_index_directory_durability.py` — 15 tests covering the flush
      ordering, both failure classifications, the digest policy under each outcome, the suppression
      scopes and their clearing, and the two crash-adjacent states (a stale cross-host reader still
      refused; a lost unflushed rename refusing loudly and naming the repair).
- [x] **Mutation testing:** 16 mutations of the changed logic, all killed by tests that name them.

The new tests state in their own docstrings that they pin the *mechanism* and cannot demonstrate
durability — that needs power loss or a crash-consistency harness. The crash tests replay the resulting
state, not the crash.

## 🏗️ Checklist

- [ ] I have made corresponding changes to the documentation — none needed; the behaviour is internal to
      index writing and surfaces only as operator warnings.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved index update reliability by ensuring committed digests are recorded even when directory durability checks are unavailable or refused.
  * Added recovery handling for stale reads and crash-related index states.
  * Improved warnings for degraded durability, avoiding repeated notifications while preserving separate filesystem and directory failure details.
* **Tests**
  * Expanded coverage for atomic writes, directory flushing, failure scenarios, stale data rejection, recovery messaging, and warning behavior.

### DIFF
--- a/src/gwmock/cli/simulate_utils.py
+++ b/src/gwmock/cli/simulate_utils.py
@@ -937,8 +937,10 @@ def _atomically_write_index(index_file: Path, index: dict[str, Any]) -> _Committ
         is that reading ``index_file`` can be served from a stale client cache, so digesting a re-read
         could record the digest of this client's *stale view* — and a later client with the same stale
         view would then match it and overwrite silently, which is the bug rather than the fix. The
-        flush outcome travels with it because recording that digest is only sound when the name it
-        describes will still be there after a crash.
+        flush outcome travels with it so the caller can *report* a rename whose durability is
+        unverifiable. It is not a gate on recording the digest: an earlier version of this function made
+        it one, and that turned a rare loud failure into a routine silent one -- see the decision comment
+        in :func:`update_signal_index`.
 
     Raises:
         OSError: If the file cannot be written or renamed.
@@ -1137,7 +1139,8 @@ def _fsync_directory(directory: Path) -> _DirectoryFlush:
         directory: Directory whose entries should be flushed.
 
     Returns:
-        Which of the three outcomes occurred.
+        Which of the four outcomes occurred: flushed, unsupported by the platform, refused by the
+        filesystem, or a directory that could not be opened.
     """
     # No `pragma: no cover`. Every branch here is reached by
     # `tests/cli/test_index_directory_durability.py` -- this one by deleting the attribute, the two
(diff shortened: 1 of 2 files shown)


isaac-cf-wong:
@coderabbitai review

PR #337, opened 2026-08-08
fix(cli): cover the population file's content in the run's identity

isaac-cf-wong:
## 📝 Summary

`run_fingerprint` identified a run by its config *bytes* plus the resolved output and metadata
directories. A config names its population catalogue by **path**, so rewriting that file changed none of
them: a checkpoint was accepted, the batches it recorded were skipped, and one run ended up holding some
batches generated from the old catalogue and the rest from the new one.

Measured before fixing, through the CLI: a 3-batch run interrupted after the first batch, the CSV
rewritten at the same path, the identical command resumed. Batch 0 kept the old catalogue's event
(detector-frame mass 30) while batches 1 and 2 carried the new one's (81, 82). **Exit code 0, no
warning.** The resume said so in its own log — `Loaded checkpoint: 1 batches already` → `Skipping batch
0`. Against this change the same reproduction refuses with `ForeignCheckpointError` and only the first
catalogue remains on disk.

The fingerprint now includes a digest of each population file the plan's batches name, de-duplicated and
sorted so adding a batch does not change the input part of the identity.

**What is hashed depends on what the loader parses.** For a `.csv`, line endings are unified and trailing
blank lines dropped: the run consumes the *parsed* catalogue, population files are machine-generated, and
a regeneration differing only in CRLF/LF would otherwise refuse a resume whose output is identical. An
interior lone `\r` is left as content — inside a quoted field it is data. For `.hdf5`/`.h5` the bytes are
hashed exactly as they are, because there `\r` and `\n` are values, not line endings. The suffix set
mirrors `gwmock_pop.loaders.file_loader.infer_population_file_format`, and a test asserts agreement with
it whenever that helper can be imported.

**Two gaps are reported rather than closed.** A remote population is not fetched to identify a run, so
its content is not covered; `report_unverified_inputs` says so on a resume at `INFO`, naming the
catalogue and the remedy. It is `INFO` and not `WARNING` deliberately: the check cannot distinguish a
commit-pinned URL — where mixing cannot occur, and what the examples use — from a mutable one, so warning
would fire on every remote resume with no way to silence it by following its own advice. An unreadable
local file *does* warn, because staging it or fixing its mode makes the check work.

Scope is the population file only. Other paths a config can name (PSD files, waveform tables) remain
uncovered, and the docstring says so instead of implying otherwise.

Tier: **T2** (behaviour).

## ✨ Type of Change

- [ ] 📝 Documentation update

- [x] **Unit tests:** `uv run pytest` — **1176 passed** with the `jax` extra, **1134 passed** without it
      (CI's plain `uv sync --group dev` cell).
- [x] **Regression test confirmed failing on the unfixed code**, for the defect: `DID NOT RAISE
      ForeignCheckpointError`. It derives its checkpoint from a genuinely interrupted run rather than a
      hand-written one — the first version failed with a `TypeError` from an argument that did not exist
      yet, which proves nothing.
- [x] **New tests:** `tests/cli/test_run_identity_covers_the_population_file.py` — 35 tests covering the
      resume refusal, the fingerprint's inputs, `~` expansion, the URL predicate's agreement with the
      loader, the text/binary split, every line-ending form that parses alike, the content changes that
      must still differ, the chunk-boundary cases, and both report levels.
- [x] **Mutation testing:** 17 mutations of the changed logic, all killed.

**Cost, since this is on every run's critical path:** 18 ms for a 45 MB (one-million-row) catalogue,
≈2.5 GB/s, anchored against `openssl speed -evp sha256` reporting 2.75 GB/s on the same host. That is a
local-SSD, CPU-bound figure; on a shared cluster filesystem a cold read of that file is I/O-bound and
plausibly ~1 s. Negligible either way against a run measured in minutes.

## 🏗️ Checklist

- [ ] I have made corresponding changes to the documentation — none needed; the behaviour surfaces as a
      refused resume and two log lines.

## Summary by CodeRabbit

* **Bug Fixes**
  * Checkpoint resuming now detects changes to referenced population files and prevents unsafe resume operations.
  * Equivalent CSV formatting changes, such as line-ending differences, no longer cause unnecessary checkpoint mismatches.
  * Resume operations now clearly report inputs that cannot be verified, including remote or unreadable files.
  * Run identity tracking now handles multiple population inputs, binary files, and reordered data consistently.

### DIFF
--- a/src/gwmock/cli/simulate_utils.py
+++ b/src/gwmock/cli/simulate_utils.py
@@ -34,7 +34,13 @@ from gwmock_noise import SimulationResult
 from tqdm import tqdm
 
 from gwmock.cli.adapter_orchestration import AdapterOrchestrationResult, AdapterOrchestrator
-from gwmock.cli.utils.checkpoint import CheckpointManager, require_matching_config, run_fingerprint, spillover_applies
+from gwmock.cli.utils.checkpoint import (
+    CheckpointManager,
+    require_matching_config,
+    run_fingerprint,
+    spillover_applies,
+    warn_if_inputs_are_unverified,
+)
 from gwmock.cli.utils.config import OrchestrationConfig, SimulatorConfig, resolve_class_path
 from gwmock.cli.utils.environment import capture_environment
 from gwmock.cli.utils.hash import compute_content_hash, compute_file_hash
@@ -2042,11 +2048,12 @@ def execute_plan(  # noqa: PLR0915
     # swapping that file's bytes left this identity unchanged and a resume mixed two catalogues into one
     # run -- measured at batch 0 holding the old catalogue's event while batches 1 and 2 held the new
     # one's, exit code 0.
+    referenced_populations = _referenced_population_files(plan)
     plan_sha256 = run_fingerprint(
         [batch.config_sha256 for batch in plan.batches],
         output_directory,
         metadata_directory,
-        _referenced_population_files(plan),
+        referenced_populations,
     )
     if checkpoint:
         require_matching_config(checkpoint.get("config_sha256"), plan_sha256, checkpoint_manager.checkpoint_file)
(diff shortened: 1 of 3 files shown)


isaac-cf-wong:
@coderabbitai review

PR #338, opened 2026-08-08
docs(cli): correct two stale comments in the index-durability code

isaac-cf-wong:
## 📝 Summary

Two stale comments in the index-durability code, both found by a reviewer while reviewing an unrelated
branch and filed separately rather than ridden in on it.

**The `_DEGRADED_TARGETS` note claimed every production caller holds the index's `flock`.** It does not:
`_exclusive_index_lock` deliberately yields **unlocked** when `fcntl` is missing, and when `flock` fails
with a locking-unsupported errno such as `ENOLCK` or `EOPNOTSUPP` — so on those filesystems two hosts
reach that set with nothing serialising them. The conclusion is unchanged (the worst case is a duplicated
or missing *warning*, never a correctness path) but the justification was false, and it is the
justification a future reader would rely on when deciding whether a lock belongs there.

**`_fsync_directory`'s docstring still said "the two failures"** after the unopenable-directory case was
split into a permission failure and everything else. There are three, and each names a different repair.

Tier: **T1** (documentation). No code semantics change — see below.

## ✨ Type of Change

- [x] 📝 Documentation update

- [x] **Full suite:** `uv run pytest` — 1177 passed, 23 skipped.
- [x] **Mechanically confirmed prose-only**, rather than asserted: parsing this branch and `main`, then
      stripping comments *and* docstrings from both, yields identical ASTs. A T1 PR whose diff touches
      code semantics is the tier's own red flag, so it is checked instead of claimed.

## 🏗️ Checklist

- [x] I have made corresponding changes to the documentation — this change *is* the documentation.
- [ ] I have added tests that prove my fix is effective — not applicable; a comment cannot be tested, and
      the AST comparison above is the evidence that nothing testable moved.

### DIFF
--- a/src/gwmock/cli/simulate_utils.py
+++ b/src/gwmock/cli/simulate_utils.py
@@ -1033,10 +1033,17 @@ def _atomically_write_index(index_file: Path, index: dict[str, Any]) -> _Committ
 # success already covers the case where the new mount works. Accepted, and recorded here rather than
 # left for the next reader to rediscover.
 #
-# Not synchronised. Every production caller holds the index's `flock`, the writer is a sequential
-# batch loop, and the resource-monitor thread never touches it; the check-then-add is formally racy
-# but a 64-thread stress probe produced exactly one warning, and the worst case is a duplicated or
-# missing *warning*, never a correctness path.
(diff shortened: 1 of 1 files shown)


isaac-cf-wong:
@coderabbitai review

PR #339, opened 2026-08-09
test(e2e): record the compute device so one reference gates both

isaac-cf-wong:
## 📝 Summary

The end-to-end references are written on a CPU, and the entries that generate through JAX
(`execution: batched`, the ripple backend, the CW branch) run on a GPU wherever one is present. Nothing
checked that the same references still hold there, and nothing recorded which device produced them.

**The device is now part of the fingerprint.** Without it a CPU and a GPU run on one host compared as the
same environment, so the bit-mismatch note blamed "references written somewhere subtly different" for
what was simply the other device. `same_environment` compares whole fingerprints, and the eight stored
references record `device: cpu`.

That label is verified rather than assumed: a Linux/x86_64/py3.12 CPU run reproduces all eight
**bit-for-bit**, which a GPU-written reference could not do.

`_jax_device()` reports the *device*, not whether JAX is installed. With JAX absent the numerics still run
on the CPU — the same device these references were written on — so reporting "none" made CI's no-`jax`
cell read as a foreign environment and silenced the note for entries that never touch JAX.
`"unavailable"` remains its own state: a CUDA plugin that raises is a broken environment, not a CPU-only
one, and a real host produced exactly that (CUDA jaxlib installed, cards below the compute capability
current JAX supports).

Tier: **T2** — this decides what the reference comparison accepts.

## ✨ Type of Change

- [ ] 📝 Documentation update
- [x] 🎨 Refactoring (no functional changes, no API changes) — test infrastructure and its documentation

- [x] **Unit tests:** 1189 passed with the `jax` extra, 1147 without it (the cell the `_jax_device` change
      concerns).
- [x] **New tests:** `tests/e2e/test_fingerprint.py`, 12 unit tests. The inert-key defect this branch
      fixes had no test at all; three mutations are killed — the comparison reverting to
      stored-keys-only, the device dropped from the record, a missing fingerprint counting as a match.
- [x] **On a CUDA host** (RTX 5060 Ti, compute capability 12.0, via HTCondor): all eight stored references
      matched, `argmax` included, which is compared exactly. Zero bit-mismatch notes, which is what
      whole-fingerprint comparison against `device: cpu` references should produce.
- [x] **The device is genuinely in the numerical path**, measured rather than assumed: the same entry
      generated on one host in two fresh processes, `JAX_PLATFORMS=cpu` against the GPU default, produces
      3 of 3 output frames differing.

**Where the gate sits**, measured on both sides:

| difference | relative | against the 1e-06 gate |
| --- | --- | --- |
| CPU to GPU, one Linux host | 3.16e-13 | passes, 3e6 to spare |
| Linux/x86_64/py3.12 to Darwin/arm64/py3.13 | 4.17e-06 | fails, by 4x |

So the tolerance admits a device change and rejects a platform one. **What this does not establish:** that
a GPU-specific regression smaller than ~1e-06 would be caught — seven orders above the device difference.
A direct pairwise comparison would close that; it needs a process per device and cannot run in CI, so it
is tracked separately rather than bolted on here.

## 🏗️ Checklist

- [x] I have made corresponding changes to the documentation — the module docstring records the procedure
      and the measurements.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation of generated reference results across CPU and GPU environments.
  * Prevented comparisons when environment fingerprints are missing or differ, reducing misleading validation results.
  * Added clearer handling for unsupported or unavailable backend detection, including known macOS and CI limitations.

* **Tests**
  * Expanded coverage for environment fingerprint matching, device detection, and reference-result comparisons.
  * Updated reference metadata to include the execution device for more reliable validation.

### DIFF
--- a/tests/e2e/matrix.py
+++ b/tests/e2e/matrix.py
@@ -73,9 +73,9 @@ E2E_MATRIX: tuple[MatrixEntry, ...] = (
             "`execution: batched` -- a segment's events generated in one call through "
             "gwmock-signal's batched entry point, then converted back to per-event chunks. "
             "Runs on whatever JAX backend is present, so this entry is CPU on a machine without one. "
-            "The references are device-independent by measurement, not by assumption: the whole matrix "
-            "was replayed on a CUDA host (RTX 5060 Ti) against these CPU-written references and every "
-            "entry passed, `argmax` included, which is compared exactly. Needs `ripplegw`"
+            "The references are device-independent by measurement, not by assumption: all eight stored "
+            "references were replayed on a CUDA host (RTX 5060 Ti) and matched, `argmax` included, which "
+            "is compared exactly. Needs `ripplegw`"
         ),
         requires=("ripplegw",),
     ),
--- a/tests/e2e/matrix.py
+++ b/tests/e2e/matrix.py
@@ -72,8 +72,10 @@ E2E_MATRIX: tuple[MatrixEntry, ...] = (
         covers=(
             "`execution: batched` -- a segment's events generated in one call through "
             "gwmock-signal's batched entry point, then converted back to per-event chunks. "
-            "Runs on whatever JAX backend is present, so this entry is CPU here; measured on an "
-            "RTX 2080 Ti, GPU and CPU agree far inside the gate. Needs `ripplegw`"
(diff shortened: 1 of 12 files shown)


isaac-cf-wong:
@coderabbitai review

